PR #12, opened 2016-07-10
Record Update documentation

Fresheyeball:
Got confused as a newcomer to postgrest on how to update a single record. I feel the docs could benefit by having an example of updating that uses filtering. This is just proposed text.


begriffs:
Thank you very much. Docs need a lot of help. Feel free to send more fixes!

PR #24, opened 2016-11-18
corrected url which calls jwt_test

schmittr:
small correction


begriffs:
Right you are, thanks!

PR #30, opened 2016-12-26
Suggest swap file on systems with under 1GB of ram

jmealo:
Builds of 0.4 repeatedly fail on the first dependency when building on a Digital Ocean Ubuntu 16.04 box with 1GB of ram. Adding 2GB of swap fixed the issue.


begriffs:
Thanks, good point.

Issue #11, opened 2016-07-06
Document connect strings for older versions of libpq


leondutoit:
Hi

I used the centos 6 build for release 0.3.2.0 on a RHEL 6 machine and started postgrest in the same way as the documentation describes (https://github.com/begriffs/postgrest-docs/blob/master/docs/install/server.md#running-the-server). 

Doing this, however, does not work. The service starts but as soon as one does a request it fails to connect to the db and returns a rather cryptic error message: `{"details":"missing \"=\" after \"postgres://db_owner@db-host:5432/db_name\" in connection info string\n","code":"","message":"Connection error"}`. 

After some discussion with @ruslantalpa on gitter he suggested that I use a different form of connection string. Turns out that passing the connection string like this `"host=hostname user=username port=5432 dbname=dbname"` to postgrest works fine.

I would like to do a pull request to make other users aware of this detail for centos 6 builds but I am not sure where this should fit into the documentation. One possibility would be to add a small subsection or a note in https://github.com/begriffs/postgrest-docs/blob/master/docs/install/server.md#running-the-server.

If you would like me to add this to the docs please let me know where.

cheers
Leon

eradman:
The subject of this ticket should probably be "Document connect strings for PostgreSQL 8.4". The URL format for connection strings has been supported since 9.0, which was released a long time ago (September of 2010).

leondutoit:
I don't agree with the change in the title - I am using PostgreSQL 9.4. I don't think this is about the db version.

eradman:
I apologize if I misunderstood the problem. Can you try explaining again why the URL syntax didn't work? As far as I know there are no connection string options that cannot be expressed as a URL.

leondutoit:
No worries :) I don't have a deep enough understanding of the source or the build environment (shared libraries) on Centos 6 to give an accurate diagnosis.

Everything that I observed, in terms of error messages, is documented in the original issue description. I was checking the PostgreSQL server logs at the time and didn't get anything from there - I was not reaching the server so I _assume_ the error message was generated by postgrest. 

Could it be that building postgrest on Centos 6 makes it depend on older versions of libpq?

eradman:
That is possible. What does `ldd` say about which client library is used?

    $ ldd postgrest | grep libpq
        libpq.so.5 => /usr/pgsql-9.5/lib/libpq.so.5 (0x00007fcb4e33d000)

leondutoit:
[...]

PR #31, opened 2017-01-23
Document connect strings for older versions of libpq (#11)

leondutoit:
I've added a small comment where I think it will be useful. Hope this covers it :)


begriffs:
This is great, thank you very much!

PR #32, opened 2017-01-26
Fix the plurality header docs for v0.4

uniphil:
The current documentation reflects the headers that must be set for singular object responses in postgrest v0.3. This change updates the documentation to reflect the way it works in v0.4, as noted by @begriffs in https://github.com/begriffs/postgrest/issues/776#issuecomment-275026837

I hope I got it right :)

### DIFF
--- a/admin.rst
+++ b/admin.rst
@@ -248,7 +248,7 @@ As discussed in `Singular or Plural`_, there are no special URL forms for singul
 .. code:: http
 
   GET /people?id=eq.1
-  Prefer: plurality=singular
+  Accept: application/vnd.pgrst.object+json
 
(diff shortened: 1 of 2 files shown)


begriffs:
Hey thank you that's really nice. I had that on my todo list and now I can cross it off. :)

There is also one other place in the docs that needs updating: http://postgrest.com/en/v0.4/admin.html#alternate-url-structure

uniphil:
Aha! I knew it was in another place, but I didn't think to look for it in `admin`. It should be updated now :)

begriffs:
Thanks again for jumping in and fixing it.

uniphil:
You're welcome! Thanks for the quick feedback and merge :) 🍰

PR #35, opened 2017-02-10
fix: fix server-proxy-uri config

hugomrdias:


begriffs:
Thanks for jumping in and submitting a fix!

hugomrdias:
no problem hope i can give back even more at least with docs (Haskell isn't my thing :) ). We are deploying to production 2 instances of postgrest next week at my company, so we have some knowledge we can share like ansible roles, nginx confs etc.

PR #36, opened 2017-02-14
fix a missing infinitive (typo) in intro.rst

tonyatoms:


begriffs:
Nice, a translation from caveman to English. 😀 

Thanks for jumping in with a fix.

PR #43, opened 2017-02-20
replace code:postgres by code:sql

ppKrauss:
Code blocks with code:postgres are not highlighted (see site navigation), not works. Replaced by code:sql that works "fine"


begriffs:
Thanks!

PR #46, opened 2017-03-23
Add section for binary output

steve-chavez:
Added the docs for https://github.com/begriffs/postgrest/pull/802. Also I changed the order of the OpenAPI section, I felt it out of place somehow and thought it was better placed at the end of the page.


begriffs:
Looks good, thanks!

PR #49, opened 2017-03-28
Role information should be stored in app_metadata.

timbod7:
I'm new to auth0, but my reading of the documentation suggests that per user app_metadata is the correct place to store roles.


begriffs:
Thanks for this, I'll review the auth0 docs about it and most likely merge this change in the next few days.

begriffs:
Another PR came in with this exact same suggestion so looks like the right thing to do! I'm going to merge that one because it adds a little extra logic to guard null objects, but thank you for pointing out this vulnerability.

PR #51, opened 2017-03-30
Fix Bad SQL coding practices in "Computed Columns" example

dk1983:
Pull request for fix https://github.com/begriffs/postgrest/issues/839, props to @vog for for the updated PostgreSql index.


begriffs:
Thanks for taking the initiative on this one!

PR #54, opened 2017-04-10
api.rst typo: delting -> deleting

elimisteve:


begriffs:
Thanks for "delting" the error. 😄

PR #55, opened 2017-04-21
Auth0 -- app metadata instead of user metadata

will-ob:
From the [docs](https://auth0.com/docs/metadata):
> An authenticated user can modify data in their profile's `user_metadata`, but not in their `app_metadata`.

Seems role-type information should be stored in `app_metadata`. For s'curity


will-ob:
content-4550_1

begriffs:
As this is the second PR suggesting the change, I believe you! 😉

begriffs:
Your version looks safer with the `|| {}` so I'll go with this PR.

PR #68, opened 2017-05-16
auth.rst: minor fixes to example scripts

mattbretl:
Just some minor fixes to the example scripts to make them functional.  I'm not sure if the "jwt" schema was intentional; if so, the purpose should probably be stated in the docs.


begriffs:
You're right, when I created that example I had installed the pgjwt extension into the jwt schema by manually running the `pgjwt--0.0.1.sql` script. (I didn't want to expose functions like `sign` or `verify` by to web clients.) However the default way of creating the extension will put these functions into the main schema so your change is correct for most users.

PR #70, opened 2017-05-18
Remove function to suppress current_setting raise with optional …

michelp:
…second argument that accomplishes the same effect


michelp:
I'm not 100% on that second change to current_setting for the secret, maybe it should raise.  if jwt.sign() gets a null secret it returns null

begriffs:
Well that's certainly a lot easier, thanks! Guess I should have read the fine manual on the current_setting function. 😄

PR #77, opened 2017-05-29
api.rst: Added example alluded to in docs

elimisteve:


begriffs:
Nice. Thanks for your help all around the docs.

elimisteve:
@begriffs No prob :-). Thanks for creating PostgREST! I will never write CRUD logic in any language ever again.

PR #79, opened 2017-05-30
Address JWT security in docs, related to postgrest issue #842

leondutoit:
As discussed in https://github.com/begriffs/postgrest/issues/842


begriffs:
Thanks for this. I wasn't even aware of `alg=none` and yeah we don't accept that!

PR #87, opened 2017-07-04
intro.rst: Correct first example app link

ProbablePrime:
The first example on the doc site was redirecting to a page within the read the docs site due to the link target and text being swapped.


begriffs:
Whoops, thank you!

PR #88, opened 2017-07-07
Single quotes give an error

chrissound:
of postgrest: ParseError "postgrest.config" "endOfInput"


begriffs:
Oh you're totally right, double quotes are correct. Thank you!

PR #90, opened 2017-07-08
Add docs for binary rpc and get master up-to-date

steve-chavez:
An option for maintaining the docs could be pushing the docs for unreleased features in master and if there's a correction in latest version of the docs do the PR in the version branch then merge in master, that way master is always up-to-date.


begriffs:
I was planning to get the docs all up to date at least as of v4.1, committing to the v4.1 branch. Once it's up to date then do a hard reset of master to the final commit of v4.1, and proceed to document new v4.2 things on master. When those things look good and the new version is released then create a v4.2 branch off of master which will make it appear as a possible version in the dropdown under the table of contents on the docs.

begriffs:
Yep, exactly.

PR #91, opened 2017-07-08
Add documentation for binary output in rpc, Fix #84

### LINKED ISSUE #84: Document how to return raw data rpc call
We already document `Accept: application/octet-stream` in https://postgrest.com/en/v4.1/api.html?highlight=octet#binary-output, but we should also note how it works with RPC. (Functions returning scalars work fine, and those returning rows need to be filtered to a single column)

steve-chavez:


begriffs:
OK sorry I was being sloppy before and v4.1 was ahead of master. I have fixed things now. Could you open this PR one last time except pointing at master?

In the future how we'll do it is that all PRs will be against master, and after squashing/merging them I will cherry pick the commit into all docs versions for which it's relevant. (If the change applies to a version that is not yet released it won't be cherry picked anywhere, but later on the new version will branch off master and end up including it.)

PR #92, opened 2017-07-09
Add documentation for binary output in rpc, Fix #84

### LINKED ISSUE #84: Document how to return raw data rpc call

steve-chavez:


begriffs:
Looks good! Only thing is that the names of `closest_region` and `overlapping_regions` don't really suggest scalar vs non-scalar return values to me. I'm imagining that closest_region would return a single row but it would still have multiple columns.

steve-chavez:
See if now is more understandable.

begriffs:
Thank you!

Issue #85, opened 2017-07-02
Verify and document Auth0 integration


begriffs:
Existing information is out of date

steve-chavez:
Now that https://github.com/begriffs/postgrest/pull/1091 got merged maybe we can use the guide in https://github.com/begriffs/postgrest/issues/906#issuecomment-326074475 to provide an example for this.

steve-chavez:
A workaround for json web key sets is mentioned here https://github.com/PostgREST/postgrest/issues/1130#issuecomment-400220291

PR #99, opened 2017-08-30
Add notice about reverting OIDC default

will-ob:
Auth0 is pushing users to a new authentication process.
To this end, they've made default a setting that postgrest cannot yet accommodate. See details in begriffs/postgrest#906.

Users can continue integrating with Auth0 if they
 - Toggle off OIDC-conformant authentiction, which, as of yesterday, is default for new clients
 - Remove `audience` configuration from the call to auth0's authentication endpoint
  This configuration has been added to the example integrations

To see current broken setup:

 1. Create rule
```
function (user, context, callback) {
  user.role = 'basic_user'
  callback(null, user, context);
}
```
 2. Create new Auth0 client in dashboard
 3. Download the sample js app; `npm install`; `serve 8080`
 4. In `app.js` change in `webAuth` config:
  - Change `scope: 'openid'` to `scope: 'openid role'`
  - Change `responseType: 'token id_token'` to `responseType: 'token'`
 5. Open `localhost:8080`, login, capture `id_token` from hash, decode [here](https://jwt.io/#encoded-jwt)
 6. Verify token does not include `role`

To fix:

 1. Toggle off `OIDC-compliant` as described in PR's note
 2. In `app.js` change in `webAuth` config:
 - remove `audience` entry
 3. Refresh the page, login, capture `id_token` from hash, decode at jwt.io, verify claim `role: basic_user` is included

Doesn't solve #85, but it's a step in the right direction.

### DIFF
--- a/auth.rst
+++ b/auth.rst
@@ -204,6 +204,16 @@ An external service like `Auth0 <https://auth0.com/>`_ can do the hard work tran
 
 To use Auth0, copy its client secret into your PostgREST configuration file as the :code:`jwt-secret`. (Old-style Auth0 secrets are Base64 encoded. For these secrets set :code:`secret-is-base64` to :code:`true`, or just refresh the Auth0 secret.) You can find the secret in the client settings of the Auth0 management console.
 
+.. note::
+
+  Make sure OIDC-conformant is toggled off.
+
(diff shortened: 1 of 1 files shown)


begriffs:
Thanks for this note and helping users get by while we work on the linked issue.

PR #109, opened 2017-10-07
Document long jwt secrets as required

begriffs:
Helps with the confusion in https://github.com/begriffs/postgrest/issues/977


turbolent:
@begriffs Thank you for adding this to the docs! I just ran into https://github.com/begriffs/postgrest/issues/977 and re-read
https://postgrest.com/en/latest/tutorials/tut1.html#step-2-make-a-secret

It mentions `openssl rand -base64 32`, but there's no mention of the length requirement, and why exactly 32. Maybe also point it out on that page, as many first time users will follow the tutorial and try a shorter than 32-chars long secret.

begriffs:
@turbolent I changed the command line snippet to pull from /dev/urandom rather than the openssl stuff. It also  now has a bolded sentence saying we need at least 32 characters.

PR #113, opened 2017-11-04
Fix typo in api.rst

hardbyte:


begriffs:
Thanks, an "ex-maple" sounds like some kind of unfortunate tree... 😄

hardbyte:
No problem, I'd have left it if the example was treacle related :-P

I have only just discovered postgREST and I'm liking what I'm learning - thanks for your efforts!

PR #121, opened 2018-01-14
Changes in 4.4 release

begriffs:
This PR updates the docs to be ready for a public 4.4 version.

The topics changed or added are each in a separate commit.

Please advise if I overlooked anything.

### DIFF
--- a/api.rst
+++ b/api.rst
@@ -563,10 +563,10 @@ Stored procedures can access request headers and cookies by reading GUC variable
 
   SELECT current_setting('request.header.origin', true);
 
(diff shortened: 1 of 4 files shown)


steve-chavez:
Because of https://github.com/begriffs/postgrest/pull/978/ the parts that say `primary key of the table of the resource being embedded must be specified`/`Here it is not necessary to specify the table's primary key of the embedded resource` in https://github.com/begriffs/postgrest-docs/blob/master/api.rst#resource-embedding can be removed.

PR #125, opened 2018-03-14
Remove iAdvize from In Production list

pocman:


begriffs:
Were there particular problems with postgrest which led you to drop it? I'd like to keep in mind what aspects to improve in the future.

pocman:
Postgrest is not compatible with our current software QA process.
We had issues with : 
* headers handling (mainly authorization)
* migrations handling
* testing
* logging
* complexity induced by functions and triggers
* lack of support for prometheus metrics

begriffs:
Thanks for the notes.

PR #133, opened 2018-05-04
add triggerfs to production apps in examples

omani:


begriffs:
Just a note about the docs system, when you merge to master the change will appear on http://postgrest.org/en/latest/ and you'll need to cherry-pick the commit to v4.4 to have it appear on https://postgrest.org/en/v4.4/ as well

omani:
ok good to know. thanks for doing that already.

PR #138, opened 2018-05-16
Add volume to docker-compose

agent3bood:


steve-chavez:
Thanks for the addition @agent3bood I think is handy to note how to do this, a couple things though.

I think it'll be better to persist the volume at the current directory instead of the root, also it's better to keep the defaults where possible so better not to modify `PGDATA`, lastly I'm not sure every user would want this by default so we could keep the change as a comment.

So to summarize I think the change should be like:
```yml
# Uncomment this if you want to persist the data.
# volumes:
#   - "./pgdata:/var/lib/postgresql/data"
```
What do you think?

agent3bood:
Thanks for the comments, I have made the changes
Awesome documentations by the way.

install.rst
```diff
@@ -222,8 +222,12 @@ To avoid having to install the database at all, you can run both it and the serv
       POSTGRES_DB: app_db
       POSTGRES_USER: app_user
       POSTGRES_PASSWORD: password
+      # Uncomment this if you want to persist the data.
+    # volumes:
+  #   - "./pgdata:/var/lib/postgresql/data"
 
 Go into the directory where you saved this file and run :code:`docker-compose up`. You will see the logs of both the database and PostgREST, and be able to access the latter on port 3000.
+This will also create a volume at /data/postgres to persist data.
```
steve-chavez: Since the change is now a comment, this line is not necessary.

PR #139, opened 2018-05-16
update PGRST_DB_ANON_ROLE in docker-compose

agent3bood:
anon should not be the default db user

### DIFF
--- a/install.rst
+++ b/install.rst
@@ -212,7 +212,7 @@ To avoid having to install the database at all, you can run both it and the serv
     environment:
       PGRST_DB_URI: postgres://app_user:password@db:5432/app_db
       PGRST_DB_SCHEMA: public
-      PGRST_DB_ANON_ROLE: app_user
+      PGRST_DB_ANON_ROLE: anon
 
   db:
     image: postgres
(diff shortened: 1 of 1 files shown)


steve-chavez:
I'd also like to enforce this from the start, however after a `docker-compose up` the request to `localhost:3000` will be an error, at the `installation` phase it's better that the user can see all is working right with few steps, to do it right maybe we would have to instruct the user to create a sql file with the `CREATE ROLE anon` and mount it to `docker-entrypoint-initdb.d` but somehow feels a bit redundant because we have tutorials.

Maybe this could be a comment with a warning like:
```
PGRST_DB_ANON_ROLE: app_user #In production this role should not be the same as the one used for the connection
```

install.rst
```diff
@@ -224,6 +224,7 @@ To avoid having to install the database at all, you can run both it and the serv
       POSTGRES_PASSWORD: password
 
 Go into the directory where you saved this file and run :code:`docker-compose up`. You will see the logs of both the database and PostgREST, and be able to access the latter on port 3000.
+Make sure to create the anon role afterwards.
```
steve-chavez: Since the change is now a comment, this line is not necessary.

steve-chavez:
Check my latest comment, also see if you can `rebase` since there's a conflict, after that I'll merge. Thanks.

PR #151, opened 2018-06-09
Add MotionDynamic (PostgREST + SubZero)

FGRibreau:


steve-chavez:
Thanks @FGRibreau, updated docs already live in https://postgrest.org/en/v5.0/intro.html#in-production.

PR #159, opened 2018-07-13
Update tut0.rst

PJLindsay:
Add hint about port mapping for users with pre-existing PostgreSQL DB


steve-chavez:
Thank you @PJLindsay, just modified the indentation for a proper render in readthedocs.

steve-chavez:
Note already live in https://postgrest.org/en/v5.0/tutorials/tut0.html#step-2-install-postgresql.

PR #179, opened 2018-09-10
For v5.1

steve-chavez:
Updating docs.. work in progress.


steve-chavez:
I think this is good to merge, one thing I don't understand is why locally the code snippets show with a white background but on rtd it shows a green one.

steve-chavez:
The green messes the important/warning notes, I'm going to overwrite that style to the one I see locally.

Issue #172, opened 2018-08-16
Resource embedding on Views note


steve-chavez:
In https://postgrest.org/en/v5.0/api.html. Here we could mention things like view fk columns detection, auto-updatable views, the RLS issue with views and the workaround.

steve-chavez:
Instead of another section for views, I'll just add the view fk columns detection note on Resource Embedding.

steve-chavez:
I'll use the "source table" term for referring to the tables from which the view SELECTs FROM.

There are two terms I've seen used for this: [base table](http://www.postgresqltutorial.com/postgresql-views/) and [source table](https://docs.microsoft.com/en-us/sql/relational-databases/views/get-information-about-a-view?view=sql-server-2017).

Source table seems more appropriate to me.

Issue #81, opened 2017-06-22
Provide auth example with app users + row level security


begriffs:
This blog post presents a great approach for auth:
https://blog.2ndquadrant.com/application-users-vs-row-level-security/

Postgrest does not allow arbitrary user queries, so the article's approach becomes really simple. Our docs need a clear example of how to do auth.

PierreRochard:
This may be a little tangential... I banged my head against RLS and ended up enforcing it in the API views like so: [`WHERE admin.table_settings.user = current_user;` ](https://github.com/PierreRochard/postgrest-boilerplate/blob/master/models/api.py#L113) Is this a valid alternative or is there a vulnerability? 

I'll work through 2ndQuadrant's blog post this weekend to contrast

begriffs:
Note to self, also include example of read-only or invisible fields based on user permissions as @sqwishy requested in the chat.

begriffs:
Also @daurnimator poses another scenario we might want to model:

> you can see all friends details; some friends of friends details; and anyone else visibility is determined by privacy settings

begriffs:
Planning to make this tutorial 2 (we've got 0 and 1 already written)

lidorcg:
Hi, I'm trying to tackle authentication and I followed [the example on the docs](https://postgrest.com/en/stable/auth.html#sql-user-management) and it seems I have generated the mentioned functions and tables but:

1. I can't reach the login function on /rpc/login
2. The user registration part is not the clearest to me - I think it's accomplished by insertion to the user table, but the auth schema is not published so I'm not sure what is the best way to solve this.

If I could get some pointers on how to continue, I'd also be very happy to contribute to the docs.

Thanks :)

begriffs:
Sorry for the silence. Did you manage to make the login function work?

JacobEvelyn:
I too am struggling with this example. First I was getting

```
ERROR:  type "basic_auth.jwt_token" does not exist
```

when trying to create the `login` function. Then when I changed that to `public.jwt_token` I can't seem to execute the login function with `/rpc/login` as @lidorcg described.

It feels like I'm close but I'm pretty inexperienced with functions/stored procedures in Postgres. 😕

lidorcg:
Sorry for not replying as well, I haven't been able to solve the problem myself (probably for lack of trying).
However, I have found the [subzero project](https://github.com/subzerocloud/subzero-starter-kit) which has users and authentication procedures built-in.

[...]

PR #181, opened 2018-09-23
Note Postgres' (current) RLS limitations for views

kyle-johnson:
This is a pretty major caveat for views which ought to be noted (see #172).

I did not  note the read-only workaround (create a stored procedure with the "SECURITY INVOKER" option).

Relevant chatter about this:
- https://stackoverflow.com/a/33863371
- https://www.postgresql.org/message-id/flat/CABFpK629xHR-jbgO0tiprMr9aZLiofMkQ3E7QxqHFRVFNcqnSQ%40mail.gmail.com

### DIFF
--- a/api.rst
+++ b/api.rst
@@ -98,6 +98,10 @@ The view will provide a new endpoint:
 
   GET /fresh_stories HTTP/1.1
 
+.. important::
+
+  Views bypass all row-level security features and are invoked as the role which created the view, much like stored procedures with the "SECURITY DEFINER" option.
+
 .. _fts:
 
 Full-Text Search
--- a/api.rst
+++ b/api.rst
@@ -100,7 +100,7 @@ The view will provide a new endpoint:
 
 .. important::
 
-  Views are invoked with the privileges of the view owner, much like stored procedures with the "SECURITY DEFINER" option. When created by a SUPERUSER role, all row-level security will be bypassed unless a different owner is specified.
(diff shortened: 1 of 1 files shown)


steve-chavez:
Thank you! The `SECURITY DEFINER` analogy is really apt.

Note, however, that RLS will only be bypassed if the role that owns the view has SUPERUSER or BYPASSRLS privileges, so views can work with RLS if we define a less privileged role for them, check my comment https://github.com/PostgREST/postgrest-docs/issues/81#issuecomment-385497947 for the workaround.

ruslantalpa:
i.e. the text in this PR is not correct, it implies RLS does not work for views. Closer to the trough would be text like: "in order for RLS to be applied/used, the owner of the view can not be a SUPERUSER" 

This behaviour is logical and the same thing happens when you try to "observe RLS in action" and connect to the database as a superuser.

kyle-johnson:
Tweaked:

"Views are invoked with the privileges of the view owner, much like stored procedures with the "SECURITY DEFINER" option. When created by a SUPERUSER role, all row-level security will be bypassed unless a different owner is specified."

kyle-johnson:
Thanks, that's detail I wasn't aware of. Changed to:

"Views are invoked with the privileges of the view owner, much like stored procedures with the "SECURITY DEFINER" option. When created by a SUPERUSER role, all row-level security will be bypassed unless a different, non-SUPERUSER owner is specified."

PR #184, opened 2018-10-15
Move intro to index and add shields

steve-chavez:
Refining the intro page, preview at http://postgrest.org/en/preview/


steve-chavez:
I think it's an improvement, I have more ideas to refine the intro page but for now I'll just deploy this one.

PR #185, opened 2018-10-19
Add description and example for get binary data with output parameter…

achubutkin:
… in stored procedure.

Extend description of Binary Output section. Add description and example for get binary data with output parameter in stored procedure. Current example without select parameter in query and scalar procedure is`t working. 

Please, check it.

### DIFF
--- a/api.rst
+++ b/api.rst
@@ -403,6 +403,17 @@ for this case.
 
   POST /rpc/closest_point HTTP/1.1
   Accept: application/octet-stream
+  
+Or with output parameter in stored procedure. You need to do a :code:`select` in this case for GET binary output.
+
+.. code-block:: postgres
+
+  CREATE FUNCTION closest_point(output _out1 bytea) AS $BODY$ BEGIN SELECT _col1 INTO _out1 FROM t_table1 END; $BODY$;
+
+.. code-block:: http
+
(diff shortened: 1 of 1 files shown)


steve-chavez:
Thanks for jumping in to contributing to the docs.

Now, for a single OUT param a `?select` shouldn't be needed(we have a [test for this](https://github.com/PostgREST/postgrest/blob/master/test/Feature/RpcSpec.hs#L273-L275)) if you reload postgrest schema cache, see http://postgrest.org/en/v5.1/admin.html#schema-reloading. Try creating the function and restarting postgrest to see how it works. 

We could definitely use an example with an OUT param in the docs though. Probably in the scalar functions section http://postgrest.org/en/v5.1/api.html#scalar-functions.

achubutkin:
Thank you so much for your answer! We tried to do as written in the documentation and encountered an error. The scalar function does not return a value.

Only case with OUT parameter and `?select` for get this parameter works for us.

steve-chavez:
Did you make sure to restart PostgREST after creating the scalar function and then repeating the request without `select`? 

Also could you share a reproducible test case(function sample plus how you make the request and the error you get)? It's likely a bug if it indeed does not work without `select`.

PR #188, opened 2018-11-17
Update to include Heroku deploy steps

lee-40square:


install.rst
```diff
@@ -336,6 +336,19 @@ If you want to have a visual overview of your API in your browser you can add sw
 
 With this you can see the swagger-ui in your browser on port 8080.
 
+Deploying to Heroku
+===================
+Assuming your making modifications locally and then pushing to GitHub, it's easy to deploy to Heroku.
+ 
+1. Create a new app on Heroku
+2. In Settings add the following buildpack :code:`https://github.com/PostgREST/postgrest-heroku`
+3. Add the require Config Vars in Heroku (see above)
```
steve-chavez: Instead of (see above) maybe we could add the list of vars we have in https://github.com/PostgREST/postgrest/blob/master/app.json#L7-L57

install.rst
```diff
@@ -336,6 +336,19 @@ If you want to have a visual overview of your API in your browser you can add sw
...
+4. Modify your postgres.conf file as required to match your Config Vars in Heroku
```
steve-chavez: I think it'd be better call the config `postgrest.conf`(extra `t`) we call it like that in other places in the docs.

install.rst
```diff
@@ -336,6 +336,19 @@ If you want to have a visual overview of your API in your browser you can add sw
...
+5. Create your :code:`Procfile` and add :code:`./env-to-config ./postgrest postgres.conf`
```
steve-chavez: Ditto. `postgrest.conf`.

install.rst
steve-chavez: Sorry, I meant to add a link to that list, in reStructuredText this is done with:
```
`list of config vars <https://github.com/PostgREST/postgrest/blob/master/app.json#L7-L57>`_
```

steve-chavez:
@lee-40square Thanks a lot for this, I'm not a heavy user of Heroku so I really appreciate your help.

install.rst
lee-40square: Is this what you had in mind?

lee-40square:
No problem. Every little bit helps. Is there anything more you need from me or is this good to go?

steve-chavez:
All good, thanks again! Merging :smiley:.

PR #189, opened 2018-11-25
Add alex-tan/postgrest-queries client side library

alex-tan:


steve-chavez:
Thank you :)! Already live at http://postgrest.org/en/v5.1/#client-side-libraries

alex-tan:
👍 awesome, thanks!

PR #204, opened 2019-03-09
clarify the usage of upsert when the table uses a surrogate key

paulovieira:
This problem has already been discussed in https://github.com/PostgREST/postgrest/issues/1118. I think the documentation would benefit from that knowledge, as it's a very common scenario.


steve-chavez:
@paulovieira Thank you! The clarification is helpful.

Changes already live at http://postgrest.org/en/v5.2/api.html#upsert.

PR #207, opened 2019-04-13
SQL inside pgjwt

kraserge:
Correct url https://github.com/michelp/pgjwt/blob/master/pgjwt--0.1.0.sql


steve-chavez:
@kraserge Thank you!

Change already live at http://postgrest.org/en/v5.2/auth.html#jwt-from-sql.

PR #215, opened 2019-05-30
add Datrium client library and testimonial

angarg:
we've been using PostgREST in internal production
infrastructure at Datrium for a few years now.
so, we're giving thanks in the form of a testimonial,
and providing a link to a python-based postgrest client
we developed and use.


steve-chavez:
@angarg Looks great! Thank you for taking the time! :smiley:

PR #236, opened 2019-07-23
Updated Docker section

stefan8888:
On macOS, it is also necessary to add the IP address in the listen_address within pg_hba.conf (tested with PostgreSQL 11.3 and postgREST v6.0.0)

### DIFF
--- a/install.rst
+++ b/install.rst
@@ -292,11 +292,16 @@ The database connection string above is just an example. Adjust the role and pas
 
     sudo ifconfig lo0 10.0.0.10 alias
 
-  You should then use 10.0.0.10 as the host in your database connection string. Also remember to include the IP address in the :code:`listen_address` within postgresql.conf and pg_hba.conf. For instance:
+  You should then use 10.0.0.10 as the host in your database connection string. Also remember to include the IP address in the :code:`listen_address` within postgresql.conf. For instance:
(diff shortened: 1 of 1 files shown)


install.rst
```diff
@@ -292,7 +292,7 @@ The database connection string above is just an example. Adjust the role and pas
 
     sudo ifconfig lo0 10.0.0.10 alias
 
-  You should then use 10.0.0.10 as the host in your database connection string. Also remember to include the IP address in the :code:`listen_address` within postgresql.conf. For instance:
+  You should then use 10.0.0.10 as the host in your database connection string. Also remember to include the IP address in the :code:`listen_address` within postgresql.conf and pg_hba.conf. For instance:
```
steve-chavez: Could be confusing since there's no `listen_addressses` in pg_hba.conf.

What's the entry you added in your pg_hba.conf?

```
host    all             all             10.0.0.10/0          md5
```

Something like that? Maybe we could be more precise and add that to the Note.

stefan8888: I completely agree, sorry my bad. Under _IPv4 local connections_ I added the following entry:

    host    all             all             10.0.0.10/32            trust

steve-chavez: Looks good now!

PR #239, opened 2019-08-01
Docs for 6.0

steve-chavez:
Work in progress..


steve-chavez:
Note to self: mention ReDoc in OpenAPI section.

steve-chavez:
I'll remove the TimescaleDB tutorial. I think some of the content can later be included in a more generic how-to about postgresql extensions

steve-chavez:
I think I'll not include the `root-spec`(https://github.com/PostgREST/postgrest/pull/1317) config in the docs yet since we don't have a default base spec to offer. Once we have it I'll include it.

steve-chavez:
Added the raw-media-types reference as it would be on `6.0.3`, with comma separators instead of being a list.

PR #244, opened 2019-08-15
Fix miscellaneous Sphinx warnings

erwanthomas:
Running `sphinx-build` (via the `make` targets at least) generates some minor warnings.

The PR fixes them, one per commit. Each warning is mentioned in the commit message.

Note: I couldn't find any contributing guideline, so indications are welcome.

### DIFF
--- a/tutorials/tut1.rst
+++ b/tutorials/tut1.rst
@@ -151,8 +151,10 @@ Go back to jwt.io and change the payload to
 
   {
     "role": "todo_user",
-    "exp": "<computed epoch value>"
+    "exp": 123456789
   }
+
+**NOTE**: Don't forget to change the dummy epoch value :code:`123456789` in the snippet above to the epoch value returned by the psql command.
 
 Copy the updated token as before, and save it as a new environment variable.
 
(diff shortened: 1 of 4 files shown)


tutorials/tut1.rst
```diff
@@ -151,7 +151,7 @@ Go back to jwt.io and change the payload to
 
   {
     "role": "todo_user",
-    "exp": <computed epoch value>
+    "exp": "<computed epoch value>"
```
erwanthomas: I guess it isn't a valid fix, since the epoch value is supposed to be an integer. If you're interested, I could reword the snippet and its explanation using a dummy integer value instead.

steve-chavez: I think that would be good. And idea would be to change it to something like `"<computed epoch value> make sure to remove the quotes and make this an integer"`. The least confusing way would be the best. Too bad json doesn't support comments.

steve-chavez:
@erwanthomas Thanks! 

See my review. I'll merge after we refine the "computed epoch value".

tutorials/tut1.rst
erwanthomas: Please see https://github.com/PostgREST/postgrest-docs/pull/244/commits/eac285d751abe9218d5a1f9ca955dae1924c7ad4

steve-chavez:
@erwanthomas Great PR with a good commit message. Thank you.

PR #254, opened 2019-09-16
Add server proxy uri env var for smoother swagger-ui  use with docker-compose

Rested:
When using swagger-ui with postgrest in docker compose, the host reported by the OpenAPI spec will default to `0.0.0.0` since this is the host on the container network. Since this is what swagger-ui uses for all but the first request, (to fetch the spec). This means that whenever `0.0.0.0` doesn't resolve to localhost on the host machine swagger-ui won't work. Setting `PGRST_SERVER_PROXY_URI` (server-proxy-uri) means a sensible default is set to make it more plug and play.


steve-chavez:
That's a good default, I agree. Change already live at http://postgrest.org/en/latest/install.html#containerized-postgrest-and-db-with-docker-compose.

PR #267, opened 2019-10-22
fix api.rst other status from 500 to 400

fortyanov:
_         -> HT.status400
When I have got  23514 error code from Postgres, I got 400 status code not 500.


steve-chavez:
@fortyanov This is true. Thanks for the correction.

Though I don't think we should return 400 for all cases, for example `23P01 exclusion_violation` should return a more precise `409 Conflict`. Maybe some errors would better correspond to `500` as well.

Perhaps we can later refine the errors in postgrest, for now the change is good.

steve-chavez:
@fortyanov Yes, correct. Actually now I think 500 is good for those errors since they mean "error on our side, retry the request later". 400 is "client error, request should be fixed", which should be the case for most errors coming from pg.

PR #269, opened 2019-10-24
Scarf install instructions

aviaviavi:
Hello! This change adds install instructions for postgrest with the [Scarf](https://scarf.sh) package manager (disclaimer: I'm the author of Scarf).

I saw that postgrest is using Github Sponsors. Scarf could be an additional way to support this project, especially from those using it in a commercial setting. It will collect and provide you with usage statistics as well as offer the ability to collect payments to opt out, so anyone who installs postgrest in this way will be supporting the project in one form or another.

I'd be happy to transfer package ownership and/or co-maintain it, so no extra work required on your part. Let me know what you think!


steve-chavez:
Hey @aviaviavi.

To be honest, the telemetry features of Scarf make me nervous about recommending it to our users. I don't think I can include it in our official docs. Sorry.

PR #278, opened 2019-11-14
Specify units for the pool timeout setting

silby:


steve-chavez:
@silby Thank you! I don't know how I missed that :man_facepalming:.

PR #279, opened 2019-11-14
Add schema basic.auth for the creation of user jwt

maitrepylos:
Hello, following the following error when creating jwt users: 

SQL Error [42883]: ERROR: function gen_salt(unknown) does not exist
  Indice : No function matches the given name and argument types. You might need to add explicit type casts.
I adapted and validated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 _build
 Pipfile.lock
+.idea
--- a/auth.rst
+++ b/auth.rst
@@ -363,7 +363,7 @@ Next we'll use the pgcrypto extension and a trigger to keep passwords safe in th
   basic_auth.encrypt_pass() returns trigger as $$
   begin
     if tg_op = 'INSERT' or new.pass <> old.pass then
-      new.pass = crypt(new.pass, gen_salt('bf'));
+      new.pass = basic_auth.crypt(new.pass, basic_auth.gen_salt('bf'));
     end if;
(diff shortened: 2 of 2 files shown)


steve-chavez:
@maitrepylos Hey there. Thanks for jumping in to help the docs.

Both `crypt` and `gen_salt` are `pgcrypto` functions. They shouldn't be in the `basic_auth` schema(this should only have our custom logic) but on the schema you use to create the extensions. The latter could be `public` or a schema you especially define for extensions(like `ext`).

Most likely when you did:

```postgresql
create extension if not exists pgcrypto;
```

Your search_path didn't include "public".

maitrepylos:
This means that pgcrypto and pgjwt extension must be found in each scheme that I expose through postgreRest?

steve-chavez:
pgrypto and pgjwt shouldn't be in the same schema you expose through postgrest(the db-schema config).They should be in another schema, usually `public`.

steve-chavez:
Btw, if the extensions are created in another schema, like `ext`, this should be included in the `db-extra-search-path`.

wolfgangwalther:
Not sure if the problem was ever solved, but the docs seem alright as they are. The proposed change would only work, if the first schema in search path was `basic_auth`. But in this case the qualification for function names would not be needed. So this change can only break the example. I'm closing this PR.

PR #282, opened 2019-11-25
added documentation for server-unix-socket-mode config option

Dansvidania:


configuration.rst
```diff
@@ -30,28 +30,29 @@ The user specified in the db-uri is also known as the authenticator role. For mo
...
+server-proxy-uri        String
+jwt-secret              String
+jwt-aud                 String
+secret-is-base64        Bool    False
+max-rows                Int     ∞
+pre-request             String
+app.settings.*          String
+role-claim-key          String  .role
+raw-media-types         String
+====================    ======  =========  ========
```
steve-chavez: Somehow the table is not rendering for me locally. Try formatting it like:
```suggestion
=======================  ======  =========  ========
Name                     Type    Default    Required
=======================  ======  =========  ========
db-uri                   String             Y
db-schema                String             Y
db-anon-role             String             Y
db-pool                  Int     10
db-pool-timeout          Int     10
db-extra-search-path     String  public
server-host              String  !4
server-port              Int     3000
server-unix-socket       String
server-unix-socket-mode  String  755
server-proxy-uri         String
jwt-secret               String
jwt-aud                  String
secret-is-base64         Bool    False
max-rows                 Int     ∞
pre-request              String
app.settings.*           String
role-claim-key           String  .role
raw-media-types          String
=======================  ======  =========  ========
```

Dansvidania: I totally forgot that. I hope now it works.

steve-chavez: Looking good now.

PR #285, opened 2019-12-05
Fixed the pg_listen syntax

ycheng-kf:
The command help shows 
```
pg_listen db-url channel /path/to/program [args]
```
Putting the command in quotes actually caused 
```
Can't run killall -SIGUSR1 postgrest: No such file or directory
```
error since pg_listen interprets the entire quoted command as the executable.

### DIFF
--- a/admin.rst
+++ b/admin.rst
@@ -213,7 +213,7 @@ Then run the `pg_listen <https://github.com/begriffs/pg_listen>`_ utility to mon
 
 .. code-block:: bash
 
-  pg_listen <db-uri> ddl_command_end "killall -SIGUSR1 postgrest"
+  pg_listen <db-uri> ddl_command_end killall -SIGUSR1 postgrest
 
 Now, whenever the structure of the database schema changes, PostgreSQL will notify the ``ddl_command_end`` channel, which will cause ``pg_listen`` to send PostgREST the signal to reload its cache.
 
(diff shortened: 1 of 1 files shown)


steve-chavez:
@ycheng-kf Hmm.. both cases fail for me. What's your shell and OS?

I've tried:
```bash
./pg_listen "postgres://postgrest_test_authenticator@localhost/postgrest_test" mychannel killall -SIGUSR1 postgrest

./pg_listen "postgres://postgrest_test_authenticator@localhost/postgrest_test" mychannel "killall -SIGUSR1 postgrest"

## both fail with "Can't run killall: No such file or directory" once the NOTIFY is done
```
Using the latest pg_listen on zsh and bash.

I think what's happening here is that pg_listen now requires the full path of the program(e.g. `/bin/killall` instead of just `killall`) and maybe your shell is somehow autocompleting the full path of the program.

ycheng-kf:
@steve-chavez, you are definitely right about the full path of the program. that should probably be documented as well. I am using bash '5.0.3(1)-release' on linux and the quoted version didn't work for me.

However, this works
```
$(which killall) -SIGUSR1 postgrest
```
let me update the pr too

steve-chavez:
@ycheng-kf Sorry for the delay. Change looks good now!

PR #289, opened 2019-12-20
Changed db-uri to reference tutorial database

rapind:
Also added a note in brackets about osx homebrew installs, which has postgresql listening on 5432 by default. I had to use `db-uri = "postgres://authenticator:mysecretpassword@localhost:5432/tutorial"` to get it working with my homebrew install.

### DIFF
--- a/tutorials/tut0.rst
+++ b/tutorials/tut0.rst
@@ -159,11 +159,11 @@ Now quit out of psql; it's time to start the API!
 Step 5. Run PostgREST
 ---------------------
 
-PostgREST uses a configuration file to tell it how to connect to the database. Create a file :code:`tutorial.conf` with this inside:
+PostgREST uses a configuration file to tell it how to connect to the database. Create a file :code:`tutorial.conf` with this inside (if you installed via homebrew, the default port postgresql is listening on will be 5432 and not 5433):
 
 .. code-block:: ini
 
-  db-uri = "postgres://authenticator:mysecretpassword@localhost:5433/postgres"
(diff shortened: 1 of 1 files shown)


tutorials/tut0.rst
```diff
@@ -159,11 +159,11 @@ Now quit out of psql; it's time to start the API!
 Step 5. Run PostgREST
 ---------------------
 
-PostgREST uses a configuration file to tell it how to connect to the database. Create a file :code:`tutorial.conf` with this inside:
+PostgREST uses a configuration file to tell it how to connect to the database. Create a file :code:`tutorial.conf` with this inside (if you installed via homebrew, the default port postgresql is listening on will be 5432 and not 5433):
 
 .. code-block:: ini
 
-  db-uri = "postgres://authenticator:mysecretpassword@localhost:5433/postgres"
+  db-uri = "postgres://authenticator:mysecretpassword@localhost:5433/tutorial"
```
steve-chavez: Hey there.

The `tutorial` db is not created in previous steps of the tutorial right? So it seems this won't work.

I think that for `tutorial` to be created we'd need to change the `POSTGRES_USER` env var(according to https://hub.docker.com/_/postgres).

rapind: Oh interesting. I had assumed the docker image had a tutorial database in it so that's what I created when setting it up via homebrew. If that's not the case, then my bad, and please ignore! :)

wolfgangwalther:
I don't see any mention of installing postgres via homebrew in the tutorial. The docker command clearly maps ports 5433, so the change seems like it's not needed. The default database is postgres, so this is consistent.

Closing this, everything is alright.

Issue #275, opened 2019-11-11
Add multiple tables to the query/Nested resource embedding


maitrepylos:
In the Resource Embedding section, it is not necessary to explain how to join more than two tables.
An example is proposed by @steve_chavez, is it possible to add it in the doc?

ex =  Films->Roles->Actors

`GET /films?select=*,roles(*,actors(*))&limit=10`

steve-chavez:
@maitrepylos Sure! That example will be a good one to add.

steve-chavez:
@maitrepylos Sure! That example would be a good one to add.

PR #290, opened 2019-12-21
Add embed disambiguation section

steve-chavez:
Work in progress..

Related to https://github.com/PostgREST/postgrest/pull/1430.


steve-chavez:
Preview at https://postgrest.org/en/preview/api.html#embedding-disambiguation

ruslantalpa:
the section is good and detailed but what I think it's missing is the "base documentation"  on how embeding works. It's very very underexplained. Only after a comprehensive section like this does the "disambiguation" section has meaning and is clear why it's relevant. Something along the lines of our chat:


When you want to embed a parent relation (many 2 one) you can either:
1.specify the table name and rename the key `billing:addresses(...)`
2. specify the foreign key and rename (if needed) it `billing:billing_id(...)`
3. specify the constraint name (which you can rename) `billing(...)`

When you want to embed a child relation (one 2 many):
...

When you want to embed a many 2 many relation:
...

ruslantalpa:
also the fact that embedding is recursive/can be nested needs to be mentioned somewhere with an example

steve-chavez:
@ruslantalpa I see what you mean. That's a section I had in mind. I wanted to call it something like: `Resource embedding cardinality` and mention the `o2m/m2o/m2m` possibilities(btw we're missing an `o2o` there for completeness). I think this can be added later.

The other stuff is a section about aliasing/renaming the embedded resource, I think this could be added now.

steve-chavez:
> Resource embedding cardinality... I think this can be added later...

I mention this because I've avoided mentioning the particular case of `m2m` disambiguation with the `junction` table. Instead just gave the users the tools to disambiguate. If they get an error message it should be clear enough what to do once they read this section. Otherwise it would get too long. Not all users will hit the same edge cases and adding all possible cases would be confusing.

steve-chavez:
About recursive embeds, there's another issue for that https://github.com/PostgREST/postgrest-docs/issues/275.

steve-chavez:
I've added another issue for aliasing https://github.com/PostgREST/postgrest-docs/issues/292. It could be added separately from disambiguation.

I'd like to merge the disambiguation feature and doc now, it has been open for a while.

PR #300, opened 2020-02-03
Adds new JS client library to /ecosystem

kiwicopple:
Hey Steve, just a small addition with one of our client libraries. It doesn't like like there is any ordering on the list so I claimed pole-position. Happy to reshuffle it though.

Once again, thanks for everything!


steve-chavez:
Hey Paul. Cool library!

The ordering is fine. I'll later classify the libs by language.

PR #302, opened 2020-02-07
correction url embedded resources through join tables

wscherphof:


steve-chavez:
@wscherphof Thank you!

steve-chavez:
Oh.. the example is what one would expect but actually the last part of the paragraph says something else:

> PostgREST can also detect relationships going through join tables. Thus you can request the Actors for Films (which in this case finds the information through Roles). You can also reverse the direction of inclusion, asking for all Directors with each including the list of their Films:

I think the `You can also reverse the direction of inclusion..` part should be removed.

wscherphof:
Indeed 👍

Op za 8 feb. 2020 om 20:49 schreef Steve Chavez <notifications@github.com>

Issue #297, opened 2020-01-29
Uncertainty about section on function privileges


H20-17:
I am newish to postgresql, so if I'm wrong I'm wrong and I apologise in advance. It says at https://postgrest.org/en/v6.0/api.html#stored-procedures

> Unlike tables/views, functions privileges work as a blacklist, so they’re executable for all the roles by default. You can workaround this by revoking the PUBLIC privileges of the function and then granting privileges to specific roles:

Specifically can you point me to somewhere online that demonstrates function privileges working as a blacklist. That would imply that you could do something like

```
REVOKE EXECUTE ON FUNCTION f(int8, text) FROM banned_user;
```
I didn't think it would work because if banned_user wasn't granted any rights in the first place, the revocation will have no effect behind the scenes. The banned user will still inherit EXECUTE on the function from PUBLIC just as before. I tested this, and unless my testing was wrong things worked the way I thought they would. I think that all privileges in postgresql are of the whitelist variety.

H20-17:
Note that this may just be a terminological nitpick. In that case I just want the language revised.

H20-17:
I would prefer that it said something like

> By default, the PUBLIC role is granted the EXECUTE privilege to every newly created function and by implication, all of the other roles will inherit this privilege. If, for some particular function, you want to restrict this function's use to specific roles, you must revoke the function's EXECUTE privilege from PUBLIC. Thereafter the function will only be usable by users who have specifically been granted the function's EXECUTE privilege.

steve-chavez:
@H20-17 You're absolutely right about functions privileges not working as a blacklist. That was a bad analogy. The wording with PUBLIC you propose is much better.

If you'd like you could a PR with the change and I'll merge it.

H20-17:
@steve-chavez I am happy to make a PR but what I wrote still wouldn't be fully sufficient because it's ot the literal truth. I think the following note or footnote also has to be included somewhere

[...]

Issue #255, opened 2019-09-18
Recommend to always label the functions with the strictest volatility


steve-chavez:
Having VOLATILE functions can have performance impacts. See https://github.com/PostgREST/postgrest/issues/1378#issuecomment-532354989.

Not sure for now where would this go, perhaps a best practices page.

PR #303, opened 2020-02-11
Change to api.rst to resolve issue #297

H20-17:
I did add a footnote with some details but I can remove it if you really don't want it.


steve-chavez:
@H20-17 I've found more info about PUBLIC, see [grant on database objects](https://www.postgresql.org/docs/current/sql-grant.html#SQL-GRANT-DESCRIPTION-OBJECTS) and this 
[so question](https://dba.stackexchange.com/questions/108975/postgresql-who-or-what-is-the-public-role).

So we could say PUBLIC is `an implicit group everybody belongs to`.

api.rst
```diff
@@ -1509,3 +1507,5 @@ PostgREST translates `PostgreSQL error codes <https://www.postgresql.org/docs/cu
...
 +--------------------------+-------------------------+---------------------------------+
+
+.. [#] Since, to stop a function from being universally executable we revoke its `EXECUTE` privilege from the `PUBLIC` virtual role, it would appear that functions typically start off with the `EXECUTE` privilege granted to `PUBLIC`, but that's the apparent truth and not the literal truth. In reality a function will start off with no privileges given to any role at all, and in that case the system behaves `as though` `PUBLIC` were given the function's `EXECUTE` privilege. Therefore if you see a blank “Access privileges” field in the output of `\df+`, `PUBLIC` effectively has the function's `EXECUTE` privilege (even though it was never actually granted to `PUBLIC`). Thereafter, subsequent to any successful `GRANT` or `REVOKE` operation performed on the function, Postgresql will employ its conventional privilege system. Notably the presence of `=X/grantor` indicates that `PUBLIC` has the `EXECUTE` privilege and, conversely, the absence of `=X/grantor` will imply that `PUBLIC` does not have the `EXECUTE` privilege.
```
steve-chavez: The blank access privileges is so weird. And actually it doesn't necessarily mean PUBLIC. It can also mean only OWNER. For example I can do this:
```sql
create function f() returns text as $$ select 'f' $$ language sql;
\df+ f 
-- Access privileges  blank             
revoke EXECUTE on function f() from PUBLIC;
\df+ f 
-- Access privileges   │ "steve-chavez"=X/"steve-chavez"
revoke EXECUTE on function f() from "steve-chavez" ;
-- Access privileges   blank again

-- the owner("steve-chavez") still has the privilege to execute it
-- but not PUBLIC
begin;
set local role other;
select f();
-- ERROR:  42501: permission denied for function f
rollback;
```
Maybe some mention of blank access privileges being inconsistent can be included.

I think the footnote is valuable, the  `=X/grantor` notation for PUBLIC is really worth mentioning.

H20-17:
Thanks,
so it appears that the situation is even more convoluted then I thought.

steve-chavez:
@H20-17 Yeah, though your notes here are valuable. Maybe we can reword some parts and mention the inconsistency on the blank access privileges.

api.rst
```diff
@@ -1509,3 +1507,5 @@ PostgREST translates `PostgreSQL error codes <https://www.postgresql.org/docs/cu
...
...default privileges are set otherwise, a function will begin its life with an implicit grant of the `EXECUTE` privilege to the `PUBLIC` virtual role. The term `implicit` is being used here because, for a given function, the "Access privileges" field in the output of `\df+` will start off blank even though, at this stage of the function's life, `PUBLIC` does have the function's `EXECUTE` privilege. At that time, if you make any modification to the privileges of the function, the output of the the `\df+` command will become what is conventionally expected. Notably the presence of `=X/grantor` indicates that `PUBLIC` has the `EXECUTE` privilege and, conversely, the absence of `=X/grantor` implies that `PUBLIC` does not have the `EXECUTE` privilege. If, after that, the `EXECUTE` privilege is revoked for all roles -thereby causing the "Access privileges" field in the output of `\df+` to revert to its original blank state- only the owner role (and its inheritors) will be able to execute the function. This is an inconsistency which probably should be addressed by the Postgresql developers, since the meaning of whatever `\df+` reports should (ideally) not depend on the foregoing operations.
```
steve-chavez: Here I think it'd be better to:

- Change `PUBLIC virtual role` to `PUBLIC role` since public is effectively a role:
```sql
create role public;
ERROR:  42939: role name "public" is reserved
LINE 1: create role public;
```
- The last part starting with `This is an inconsistency..` feels a bit like bashing on pg developers. Maybe just mention it's an inconsistency on `\df+` that users should be aware of. Ideally we should report that on a pg mailing list.

steve-chavez:
What do you think about moving the `Function privileges` section to another page called Best Practices?(seems more fit there than in the API page). I'm thinking we could also solve https://github.com/PostgREST/postgrest-docs/issues/255 in that page.

H20-17:
@steve-chavez,
I've been making a fool out of myself. As I said, I was new to postgreql. It's not just functions that start out with empty "Access Privileges" fields. It's everything, including tables. I just didn't notice that because I've always assigned privileges to tables immediately after creating them.

The rule is that when the "Access privileges" field is empty, "default privileges" apply.

Default privileges are defined in [https://www.postgresql.org/docs/12/ddl-priv.html#PRIVILEGES-SUMMARY-TABLE](https://www.postgresql.org/docs/12/ddl-priv.html#PRIVILEGES-SUMMARY-TABLE).

I think we should just omit the entire footnote since were looking at behaviour that is actually described in the manual. The inconsistency you found is something that should be taken up on the devel mailing list.

steve-chavez:
@H20-17 Oh, I see. I also didn't have the default privileges part clear.

I think it's ok to remove the footnote.

H20-17:
Do you still want another page called "best practices" for this?

steve-chavez:
Just removing the footnote would be good. I'll have to think more about the Best Practices page.

steve-chavez:
If you'd like, you can add the best practices page. Just move the function privileges section to a `best_practices.rst` file and then add these lines to `index.rst`.
```patch
+.. toctree::
+   :caption: Best Practices
+   :hidden:
+
+   best_practices.rst
+
 - :doc:`Authentication <auth>`
 - :doc:`Installation <install>`
 - :doc:`Administration <admin>`
+- :doc:`Best Practices <best_practices>`
```

H20-17:
No problem. I'll try to get on it later today. Sorry for all of the barking up the wrong tree.

steve-chavez:
@H20-17 Thanks a lot for your work here. It also clarified things for me :)

Issue #233, opened 2019-07-22
Provide a How-to for getting images in <img>


steve-chavez:
See https://github.com/PostgREST/postgrest/issues/1077.

steve-chavez:
Fixed in #307.

PR #307, opened 2020-03-11
Tutorial: Providing images for <img>

pkel:
I tried to setup an endpoint for providing binary data for  tags. This is what I found out.

I need some help with the following point.
- Return 404 when the requested id is not present in table.

If we could trigger raw output via a config parameter from within the procedure, we could provide a generic file endpoint. Setting `raw-media-types= */*` is probably not a good idea. Is this worth a
feature request?

Addresses #233.

### DIFF
--- a/how-tos/providing-images-for-img.rst
+++ b/how-tos/providing-images-for-img.rst
@@ -32,7 +32,7 @@ First, we need a public table for storing the files.
    );
 
 Let's assume this table contains an image of two cute kittens with id 42.
-We can retrieve this image in binary format from our PostgREST API by requesting :code:`/files?select=blog&id=eq.42` with the :code:`Accept: application/octet-stream` header.
+We can retrieve this image in binary format from our PostgREST API by requesting :code:`/files?select=blob&id=eq.42` with the :code:`Accept: application/octet-stream` header.
 Unfortunately, putting the URL into the :code:`src` of an :code:`<img>` tag will not work.
 That's because browsers do not send the required header.
 
@@ -118,8 +118,9 @@ For production, you probably want to configure additional caches, e.g. on the re
      end
    $$ language plpgsql;
 
-With this, we can obtain the cat image from `/rpc/file?id=42`.
+With this, we can obtain the cat image from :code:`/rpc/file?id=42`.
 Consequently, we have to replace our previous rewrite rule in the Nginx recipe with the following.
 
 .. code-block:: nginx
+
    rewrite /files/([^/]+).*  /rpc/file?id=$1  break;
--- a/how-tos/providing-images-for-img.rst
+++ b/how-tos/providing-images-for-img.rst
@@ -36,30 +36,15 @@ We can retrieve this image in binary format from our PostgREST API by requesting
 Unfortunately, putting the URL into the :code:`src` of an :code:`<img>` tag will not work.
 That's because browsers do not send the required header.
 
-Luckily, we can configure our Nginx reverse proxy to fix this problem for us.
-The following recipe assumes that PostgREST is running on port 3000.
-It configures Nginx as ordinary HTTP server on port 80.
-Requests to :code:`/api/*` are forwarded to our PostgREST instance.
-Requests to :code:`/files/<id>*` are forwarded to our endpoint with the :code:`Accept` header set to :code:`application/octet-stream`.
+Luckily, we can configure our `Nginx reverse proxy <../admin.html>`_ to fix this problem for us.
+We assume that PostgREST is running on port 3000.
+We provide a new location :code:`/files/` that redirects requests to our endpoint with the :code:`Accept` header set to :code:`application/octet-stream`.
 
 .. code-block:: nginx
 
    server {
-     listen 80 default_server;
-     listen [::]:80 default_server;
-     server_name _;
-     root /var/www/html;
-     index index.html;
-     try_files $uri $uri/ =404;
-
-     location /api/ {
-       default_type  application/json;
-       proxy_hide_header Content-Location;
-       add_header Content-Location /api/$upstream_http_content_location;
-       proxy_set_header  Connection "";
-       proxy_http_version 1.1;
-       proxy_pass http://localhost:3000/;
-     }
+     # rest of reverse proxy and web server configuration
+     ...
 
      location /files/ {
        # /files/<id>/* ---> /files?select=blob&id=eq.<id>
@@ -75,9 +60,6 @@ Requests to :code:`/files/<id>*` are forwarded to our endpoint with the :code:`A
        proxy_http_version 1.1;
        proxy_pass http://localhost:3000/;
      }
-
-As you can see, we only explain the :code:`files` location.
-The reasoning for the rest of the recipe can be found `elsewhere <../admin.html>`_.
 
 With this setup, we can request the cat image at :code:`localhost/files/42/cats.jpeg` without setting any headers.
 In fact, you can replace :code:`cats.jpeg` with any other filename or simply omit it.
@@ -100,7 +82,7 @@ First, we store the media types and names of our files in the database.
 
 .. code-block:: postgres
 
-   create table public.files(
+   create table files(
      id   int primary key
    , type text
    , name text
@@ -113,8 +95,6 @@ For production, you probably want to configure additional caches, e.g. on the re
 
 .. code-block:: postgres
 
-   set search_path=public
-
    create function file(id int) returns bytea as
    $$
      declare headers text;
@@ -124,7 +104,7 @@ For production, you probably want to configure additional caches, e.g. on the re
          '[{"Content-Type": "%s"},'
           '{"Content-Disposition": "inline; filename=\"%s\""},'
           '{"Cache-Control": "max-age=259200"}]'
-         , f.type, f.name)
+         , files.type, files.name)
        from files where files.id = file.id into headers;
        perform set_config('response.headers', headers, true);
        select files.blob from files where files.id = file.id into blob;
@@ -139,7 +119,7 @@ For production, you probably want to configure additional caches, e.g. on the re
    $$ language plpgsql;
 
 With this, we can obtain the cat image from `/rpc/file?id=42`.
-Consequently, we have to replace the rewrite rule in the Nginx recipe with the following.
+Consequently, we have to replace our previous rewrite rule in the Nginx recipe with the following.
 
 .. code-block:: nginx
    rewrite /files/([^/]+).*  /rpc/file?id=$1  break;
--- a/how-tos/providing-images-for-img.rst
+++ b/how-tos/providing-images-for-img.rst
@@ -25,6 +25,7 @@ We will show how to achieve this using Nginx.
 First, we need a public table for storing the files.
 
 .. code-block:: postgres
+
    create table public.files(
      id   int primary key
    , blob bytea
--- a/how-tos/providing-images-for-img.rst
+++ b/how-tos/providing-images-for-img.rst
@@ -31,15 +31,15 @@ First, we need a public table for storing the files.
    );
 
 Let's assume this table contains an image of two cute kittens with id 42.
-We can retrieve this image in binary format from our PostgREST API by requesting ::code::`/files?select=blog&id=eq.42` with the :code:`Accept: application/octet-stream` header.
-Unfortunately, putting the URL into the ::code::`src` of an ::code::`<img>` tag will not work.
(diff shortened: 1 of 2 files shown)


steve-chavez:
@pkel Really interesting and nicely structured how-to. Thank you!

Regarding `404`, you could do that with `RAISE sqlstate PT404..`(docs on that [here](http://postgrest.org/en/v6.0/api.html#errors-and-http-status-codes)) in case the row doesn't exist.

However, I think we could do this in a way that addresses your second concern(generic solution) and without custom logic for the 404.

PostgREST master version contains this feature: [Setting headers via pre-request](http://postgrest.org/en/latest/api.html#pre-req-headers). With this, the logic could be changed to detect when a request with an `Accept: image/*` comes from, for every(or some) endpoints, and then set the corresponding headers. The 404 would come for free since we wouldn't be doing an rpc request.

Here's an example of some of the logic possibilities: https://github.com/PostgREST/postgrest/blob/master/test/fixtures/schema.sql#L1647-L1666.

For trying this now you'd have to [build from source](http://postgrest.org/en/latest/install.html#build-source).

What do you think?

pkel:
Thanks for the `RAISE sqlstate PT404` tip. That will do.

The `pre-request` options looks very cool. It probably enables a lot of interesting stuff. 
But I do not see how to apply it to the file endpoint. I want to read the return content type from the database. The type depends on the file id. I think an rpc with the id as argument is more reasonable here. Am I missing something?

Additionally, I just found out a problem with the tutorial: (my) Firefox requests images with `Accept: image/webp,*/*`. In order to provide binary data, I would have to set `raw-media-types=*/*` globally. Does this cause any side effects?

It would be perfect, if the `pre-request` function could do something like
```
perform set_config('response.raw', 'true');
```
on certain endpoints, independent of the global `raw-media-types` option.

pkel:
I think there is an issue in the `raw-media-types` logic. Firefox sends navigation requests with
```
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8
```
This accept header does not cause binary output when `raw-media-types` is set to `*/*`. Removing the `;q=0.9` parts solves the issue. I suspect a parsing bug. Maybe the `;` ends the list of media types?

steve-chavez:
> I want to read the return content type from the database. The type depends on the file id. I think an rpc with the id as argument is more reasonable here. Am I missing something?

@pkel Personally I have only returned `png` so that's why I suggested that. I think the rpc is good for the how-to. Though I guess if you really wanted to keep it generic with the `pre-request` you could do something like `CREATE VIEW "images.jpg" .. FROM images .. WHERE type = 'jpg'` and then extract the extension from the request path and add the media type on the `pre-request`. There could be other ways, but again the rpc is really good for the how-to.

 > Additionally, I just found out a problem with the tutorial: (my) Firefox requests images with Accept: image/webp,*/*. In order to provide binary data. I would have to set raw-media-types=*/* globally. Does this cause any side effects?

`*/*` looks like it may cause problems. Why instead don't you set it as:

```
raw-media-types = "image/*, image/webp"
``` 

I've tested with both Chrome and Firefox and seems to be working fine.

steve-chavez:
> It would be perfect, if the pre-request function could do something like
> perform set_config('response.raw', 'true');

You could issue a feature request in the main repo for that https://github.com/PostgREST/postgrest/issues. There we could discuss how to implement it.

steve-chavez:
@pkel You could include the nginx snippet you shared in https://github.com/PostgREST/postgrest/issues/1462#issuecomment-599042207 for finishing the how-to. 

When `raw-media-endpoints` is implemented we could change it for simplicity's sake. But I think the nginx snippet is good for now.

pkel:
> @pkel You could include the nginx snippet you shared in [PostgREST/postgrest#1462 (comment)](https://github.com/PostgREST/postgrest/issues/1462#issuecomment-599042207) for finishing the how-to.

Done. Though it's more a rewrite than it is an addition.

how-tos/providing-images-for-img.rst
```diff
@@ -0,0 +1,143 @@
+Providing images for <img>
+==========================
+
+:author: `pkel <https://github.com/pkel>`_
+
+In this tutorial, you will learn how to create an endpoint for providing images to HTML :code:<img> tags without client side javascript.
```
steve-chavez:
```suggestion
In this how-to, you will learn how to create an endpoint for providing images to HTML :code:`<img>` tags without client side javascript.
```

how-tos/providing-images-for-img.rst
```diff
@@ -0,0 +1,143 @@
...
+The general idea is to configure the reverse proxy in front of the API to set this header for all requests to :code:`/files/`.
+We will show how to achieve this using Nginx.
+
+First, we need a public table for storing the files.
+
+.. code-block:: postgres
+   create table public.files(
+     id   int primary key
+   , blob bytea
+   );
```
steve-chavez:
```suggestion
.. code-block:: postgres

   create table public.files(
     id   int primary key
   , blob bytea
   );
```
It needs a newline after `code-block` otherwise it doesn't render properly.

how-tos/providing-images-for-img.rst
```diff
@@ -0,0 +1,143 @@
...
+That's because browsers do not send the required header.
+
+Luckily, we can configure our Nginx reverse proxy to fix this problem for us.
+The following recipe assumes that PostgREST is running on port 3000.
+It configures Nginx as ordinary HTTP server on port 80.
+Requests to :code:`/api/*` are forwarded to our PostgREST instance.
+Requests to :code:`/files/<id>*` are forwarded to our endpoint with the :code:`Accept` header set to :code:`application/octet-stream`.
+
+.. code-block:: nginx
+   server {
```
steve-chavez:
```suggestion
.. code-block:: nginx

   server {
```

how-tos/providing-images-for-img.rst
```diff
@@ -0,0 +1,143 @@
...
+     try_files $uri $uri/ =404;
+
+     location /api/ {
+       default_type  application/json;
+       proxy_hide_header Content-Location;
+       add_header Content-Location /api/$upstream_http_content_location;
+       proxy_set_header  Connection "";
+       proxy_http_version 1.1;
+       proxy_pass http://localhost:3000/;
+     }
```
steve-chavez: I think all of this block is non-essential for the how-to, how about if you omit it?

how-tos/providing-images-for-img.rst
```diff
@@ -0,0 +1,143 @@
...
+
+As you can see, we only explain the :code:`files` location.
+The reasoning for the rest of the recipe can be found elsewhere_.
+
+.. _elsewhere: ../admin.html#
```
steve-chavez: Can also be done like:
```suggestion
The reasoning for the rest of the recipe can be found `elsewhere <../admin.html>`_.
```
rst is weird for links.

how-tos/providing-images-for-img.rst
```diff
@@ -0,0 +1,143 @@
...
+2.  Download requests (e.g. Right Click -> Save Image As) to :code:`files/42` will propose :code:`42` as filename.
+    This might confuse users.
+3.  Requests to the binary endpoint are not cached.
+    This will cause unnecessary load on the database.
+
+The following improved version addresses these problems.
+First, we store the media types and names of our files in the database.
+
+.. code-block:: postgres
+   create table public.files(
```
steve-chavez:
```suggestion
.. code-block:: postgres

   create table public.files(
```

how-tos/providing-images-for-img.rst
```diff
@@ -0,0 +1,143 @@
...
+   , name text
+   , blob bytea
+   );
+
+Next, we set up an RPC endpoint that sets the content type and filename.
+We use this opportunity to configure some basic, client-side caching.
+For production, you probably want to configure additional caches, e.g. on the reverse proxy.
+
+.. code-block:: postgres
+   set search_path=public
```
steve-chavez:
```suggestion
.. code-block:: postgres

   set search_path=public
```

how-tos/providing-images-for-img.rst
```diff
@@ -0,0 +1,143 @@
...
+   create function file(id int) returns bytea as
+   $$
+     declare headers text;
+     declare blob bytea;
+     begin
+       select format(
+         '[{"Content-Type": "%s"},'
+          '{"Content-Disposition": "inline; filename=\"%s\""},'
+          '{"Cache-Control": "max-age=259200"}]'
+         , f.type, f.name)
```
steve-chavez: I'm getting an error when running this function:
```
ERROR:  42P01: missing FROM-clause entry for table "f"
```
For solving this you could do:
```suggestion
         , type, name)
```

how-tos/providing-images-for-img.rst
```diff
@@ -0,0 +1,143 @@
...
+---------------
+
+PostgREST returns binary data on requests that set the :code:`Accept: application/octet-stream` header.
+The general idea is to configure the reverse proxy in front of the API to set this header for all requests to :code:`/files/`.
+We will show how to achieve this using Nginx.
+
+First, we need a public table for storing the files.
+
+.. code-block:: postgres
+   create table public.files(
```
steve-chavez: We usually don't recommend using the `public` schema because extensions are usually installed there and they can create a lot of noise. I think it'll be better to just remove the schema for the how-to.
```suggestion
   create table files(
```

how-tos/providing-images-for-img.rst
steve-chavez: Ditto, I think we shouldn't assume the schema.
```suggestion
   create table files(
```

steve-chavez:
@pkel Left some minor corrections for the how-to. Overall it looks good.

Also, add another item here so that the main page has a link to the how-to:
https://github.com/PostgREST/postgrest-docs/blob/a34353653904311fb21d3033f25ced88f07baaf2/index.rst#how-to-guides

Should be like:
```
- :doc:`how-tos/providing-images-for-img`
```

how-tos/providing-images-for-img.rst
pkel: D'accord

pkel:
Thanks for your comments! I tried to address them.

how-tos/providing-images-for-img.rst
```diff
@@ -0,0 +1,125 @@
...
+         hint = format('%s seems to be an invalid file id', file.id);
+       end if;
+     end
+   $$ language plpgsql;
+
+With this, we can obtain the cat image from `/rpc/file?id=42`.
+Consequently, we have to replace our previous rewrite rule in the Nginx recipe with the following.
+
+.. code-block:: nginx
+   rewrite /files/([^/]+).*  /rpc/file?id=$1  break;
```
steve-chavez:
```suggestion
.. code-block:: nginx

   rewrite /files/([^/]+).*  /rpc/file?id=$1  break;
```

how-tos/providing-images-for-img.rst
```diff
@@ -0,0 +1,125 @@
...
+       then return(blob);
+       else raise sqlstate 'PT404' using
+         message = 'NOT FOUND',
+         detail = 'File not found',
+         hint = format('%s seems to be an invalid file id', file.id);
+       end if;
+     end
+   $$ language plpgsql;
+
+With this, we can obtain the cat image from `/rpc/file?id=42`.
```
steve-chavez:
```suggestion
With this, we can obtain the cat image from :code:`/rpc/file?id=42`.
```

how-tos/providing-images-for-img.rst
```diff
@@ -0,0 +1,125 @@
...
+
+.. code-block:: postgres
+
+   create table files(
+     id   int primary key
+   , blob bytea
+   );
+
+Let's assume this table contains an image of two cute kittens with id 42.
+We can retrieve this image in binary format from our PostgREST API by requesting :code:`/files?select=blog&id=eq.42` with the :code:`Accept: application/octet-stream` header.
```
steve-chavez:
```suggestion
We can retrieve this image in binary format from our PostgREST API by requesting :code:`/files?select=blob&id=eq.42` with the :code:`Accept: application/octet-stream` header.
```

steve-chavez:
@pkel Looking better now! 

Check my latest comments. After that I'll merge :rocket:.

steve-chavez:
> Github's GUI is a bit confusing. There is a button "review changes". Its green color indicates that it is the only way forward, but reviewing makes no sense for the author of a PR.

@pkel Weird, I haven't seen that one yet.

I'll merge now. Thanks a lot for your contribution!

PR #308, opened 2020-03-12
Add type basic_auth.jwt_token.

Rackar:
Or the function login will show error.


steve-chavez:
There's a `CREATE TYPE jwt_token..` above in http://postgrest.org/en/v6.0/auth.html#jwt-from-sql. However it's created just for a quick test, the basic_auth schema isn't even created yet. So I think your change is good despite we'll now have some duplication.

@Rackar Thank you!

PR #310, opened 2020-03-14
redraw film.png. Do you like it?

Rackar:
film


steve-chavez:
Looks nice!

What tool are you using for generating the diagram? Or did you draw it manually?

Right now we're using https://github.com/BurntSushi/erd/. Which lets us generate the diagram based on this plain text file https://github.com/PostgREST/postgrest-docs/blob/master/erd/film.er.

Rackar:
I drew it manually in Xmind.

wolfgangwalther:
While I agree, that it does look nice and I think we all appreciate the effort - we should stick with drawings that are auto-generated from code. This is easier to maintain. One immediate example would be, that we currently have two of those drawings (according to the contents of https://github.com/PostgREST/postgrest-docs/tree/master/diagrams), but this PR only changes one of them.

I think we're good with where we're at with the drawings right now. If you have any better tooling in mind, that might create nicer output from code, we'd be happy to know about that!

Closing this PR!

PR #313, opened 2020-03-21
Fix the 'Roles for Each Web User' example

mrexodia:
I struggled for a while to make it work, but in the end found that https://www.postgresql.org/docs/9.5/sql-createpolicy.html says you have to run `ALTER TABLE ... ENABLE ROW LEVEL SECURITY;` and that appears to have fixed the issue.

Also I'm not sure the documentation about `current_setting` would be good practice. If you do not pass the `true` argument to return NULL if the setting is missing the error will be much more obvious (and this is also what is done in the [linked tutorial](https://www.2ndquadrant.com/en/blog/application-users-vs-row-level-security/)).

In my opinion `current_setting('request.jwt.claim.email')` would be better, but let me know what you think because I'm not exactly an expert.

### DIFF
--- a/auth.rst
+++ b/auth.rst
@@ -60,6 +60,8 @@ You can use row-level security to flexibly restrict visibility and access for th
     message_subject VARCHAR(64) NOT NULL,
     message_body    TEXT
   );
+  
+  ALTER TABLE chat ENABLE ROW LEVEL SECURITY;
 
 We want to enforce a policy that ensures a user can see only those messages sent by him or intended for him. Also we want to prevent a user from forging the message_from column with another person's name.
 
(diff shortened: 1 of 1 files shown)


steve-chavez:
> ALTER TABLE ... ENABLE ROW LEVEL SECURITY;

That one was missing. Thank you. The truth is that the auth section is not in tutorial form yet. So there's definitely things to be filled in. 

> In my opinion current_setting('request.jwt.claim.email') would be better, but let me know what you think because I'm not exactly an expert.

Hmm.. I think the `missing_ok` version is easier to work with. With that you can handle the error case by detecting the null. If not, I'm not sure if you can somehow CATCH the `ERROR:  42704: unrecognized configuration parameter` in plpgsql. 

Usually, getting the GUC would be done by a wrapper function such as this one: https://github.com/steve-chavez/socnet/blob/master/schema/util.sql#L11-L14

PR #314, opened 2020-03-31
mention that JSON operators work for row filtering

fiatjaf:


api.rst
```diff
@@ -204,13 +204,16 @@ You can specify a path for a ``json`` or ``jsonb`` column using the arrow operat
     { "id": 2, "blood_type": "O+", "phones": [{"country_code": "43", "number": "512-446-4988"}, {"country_code": "43", "number": "213-891-5979"}] }
   ]
 
+That also works with filters:
+
 .. code-block:: http
 
-  GET /people?select=id,json_data->phones->0->>number HTTP/1.1
+  GET /people?select=id,json_data->>blood_type&json_data->blood_type=eq."A+" HTTP/1.1
```
steve-chavez: Thanks for helping the docs!

How about changing `A+` to `A-`? Otherwise I think as it is the example might require url encoding for the plus and quotes.

You could remove the quotes after that since `-` is url safe.

fiatjaf: Ok, makes sense!

PR #315, opened 2020-04-03
Latest docs for 7.0.0

steve-chavez:
Only the latest feature regarding multiple schemas is missing: https://github.com/PostgREST/postgrest/pull/1450

Ideally a "schema design" section should be added to the best practices page. That way we ensure multiple schemas are not misused. It should be made clear that the use cases are "api versioning with schemas" and "schema-based multitenancy".


steve-chavez:
For fixing #317, I've omitted talking about pg minimum version in the tutorial(after all we use docker latest version) and instead put a link to the installation page.

Also, I'm kind of discouraging the use of the `Prefer: params=multiple-objects` because it was discovered that it had bad performance https://github.com/PostgREST/postgrest/pull/1385. I'm advising users to instead use json or array parameters in functions for bulk processing.

steve-chavez:
I've added a "schema structure" diagram: https://postgrest.org/en/preview/best_practices.html.

Doing a "schema design" section is more of a tutorial, which takes longer. So I'm going to leave it as an explanation.

steve-chavez:
Ok. This is finished. Merging..

steve-chavez:
Left a warning for not misusing the switching schemas feature: http://postgrest.org/en/v7.0.0/configuration.html#db-schema.

PR #316, opened 2020-04-13
Documentation on how to run the test suite locally

monacoremo:
Documentation on the alternative ways to run tests locally that would be added with https://github.com/PostgREST/postgrest/pull/1476 and https://github.com/PostgREST/postgrest/pull/1480.

### DIFF
--- a/install.rst
+++ b/install.rst
@@ -221,10 +221,23 @@ When a pre-built binary does not exist for your system you can build the project
 PostgREST Test Suite
 --------------------
 
-Creating the Test Database
-~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


steve-chavez:
Oh, I forgot about merging this one.

The new way of running the test database is much simpler than the docker workflow.

I think later on we could remove most of the docker wording and replace it with a simple instruction of running [test/docker-compose.yml](https://github.com/PostgREST/postgrest/blob/master/test/docker-compose.yml). Probably rename the section to testing with docker.

The `test/with_tmp_db` should definitely be the preferred way.

Adding the Nix section would also be interesting.

PR #320, opened 2020-05-01
Fix link to DigitalOcean PostgREST Contact List video

nickrsan:
Old playlist is gone - linked to first video in series instead


steve-chavez:
Thank you! Change already live at http://postgrest.org/en/v7.0.0/ecosystem.html#community-tutorials.

PR #332, opened 2020-06-12
add redux-postgrest to client-side libraries

andytango:
As per https://github.com/andytango/redux-postgrest/issues/7


steve-chavez:
Cool! Thank you!

Issue #336, opened 2020-07-09
Links to pg-safeupdate in documentation are broken


sirviot:
In documentation

    * Hardening PostgREST - Block Full-Table Operations and
    * Ecosystem - Extensions

a project named `pg-safeupdate` is referred. The project has now moved from Bitbucket to GitHub.

### Expected behavior

Links in documentation point to https://github.com/eradman/pg-safeupdate

### Actual behavior

Links point to deprecated page https://bitbucket.org/eradman/pg-safeupdate/

sirviot:
I'll update the links and make a pull request.

sirviot:
Pull request #337 is now available

PR #337, opened 2020-07-09
Fix links to pg-safeupdate 

sirviot:
Change broken Bitbucket links for pg-safeupdate to GitHub.

For issue #336


steve-chavez:
@sirviot Thanks for catching these up!

PR #338, opened 2020-07-09
Fix #334: Alter default privileges

### LINKED ISSUE #334: Incorrect instructions to alter default privileges on functions
In the Schema structure document the following is prescribed:

```
ALTER DEFAULT PRIVILEGES IN SCHEMA api REVOKE EXECUTE ON FUNCTIONS FROM PUBLIC;
```

but the PostgreSQL documentation specifically informs that this **will have no effect**. From <https://www.postgresql.org/docs/12/sql-alterdefaultprivileges.html#SQL-ALTERDEFAULTPRIVILEGES-EXAMPLES>:

> Note however that you cannot accomplish that effect with a command limited to a single schema. This command has no effect, unless it is undoing a matching GRANT:
>
> ``` 
> ALTER DEFAULT PRIVILEGES IN SCHEMA public REVOKE EXECUTE ON FUNCTIONS FROM PUBLIC;
> ```
>
> That's because per-schema default privileges can only add privileges to the global setting, not remove privileges granted by it.

To make it work as expected the `IN SCHEMA api` fragment should be removed. The possible unintended side effect is that this will change default privileges in all schemas. The only way to deal with it I can think of is to order the statements. 

Shall I prepare a PR to improve this?

Thanks for all the excellent work on PostgREST!

tad-lispy:
As discussed in the #334 issue.

### DIFF
--- a/schema_structure.rst
+++ b/schema_structure.rst
@@ -38,8 +38,8 @@ After that, you'll need to grant EXECUTE privileges on functions explicitly:
 
 .. code-block:: postgres
 
+   GRANT EXECUTE ON ALL FUNCTIONS IN SCHEMA api TO web_user;
    GRANT EXECUTE ON FUNCTION login TO anonymous;
-   GRANT EXECUTE ON FUNCTION reset_password TO web_user;
 
 Security definer
 ----------------
--- a/schema_structure.rst
+++ b/schema_structure.rst
@@ -20,11 +20,17 @@ This allows you to change the internals of your schema and maintain backwards co
 Functions
 =========
 
-By default, when a function is created, the privilege to execute it is not restricted by role. The function access is PUBLIC—executable by all roles(more details at `PostgreSQL Privileges page <https://www.postgresql.org/docs/12/ddl-priv.html>`_). This is not ideal for an API schema. To disable this behavior, you can run the following SQL statement:
+By default, when a function is created, the privilege to execute it is not restricted by role. The function access is PUBLIC—executable by all roles (more details at `PostgreSQL Privileges page <https://www.postgresql.org/docs/12/ddl-priv.html>`_). This is not ideal for an API schema. To disable this behavior, you can run the following SQL statement:
 
 .. code-block:: postgres
 
-  ALTER DEFAULT PRIVILEGES IN SCHEMA api REVOKE EXECUTE ON FUNCTIONS FROM PUBLIC;
+  ALTER DEFAULT PRIVILEGES REVOKE EXECUTE ON FUNCTIONS FROM PUBLIC;
+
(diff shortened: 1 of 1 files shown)


schema_structure.rst
```diff
@@ -20,20 +20,26 @@ This allows you to change the internals of your schema and maintain backwards co
...
 .. code-block:: postgres
 
-  ALTER DEFAULT PRIVILEGES IN SCHEMA api REVOKE EXECUTE ON FUNCTIONS FROM PUBLIC;
+  ALTER DEFAULT PRIVILEGES REVOKE EXECUTE ON FUNCTIONS FROM PUBLIC;
+
+This will change the privileges for all functions created in the future in all schemas. Currently there is no way to limit it to a single schema. The key phrase here is *in the future*. Therefore you should put this statement at the beginning of the API schema definition, and then at the end reverse it with:
+
+.. code-block:: postgres
+
+  ALTER DEFAULT PRIVILEGES GRANT EXECUTE ON FUNCTIONS TO PUBLIC;
```
steve-chavez, line 33: How about if we make the reversal a note? Like "If you'd like to reverse this, you can do ALTER..".

I think we should encourage the practice of being explicit with func privileges.
Is likely that if the user wants to use all functions unrestricted, he'll login with a superuser anyway.

schema_structure.rst
```diff
@@ -20,20 +20,26 @@ This allows you to change the internals of your schema and maintain backwards co
...
 
 See `PostgreSQL alter default privileges <https://www.postgresql.org/docs/current/static/sql-alterdefaultprivileges.html>`_ for more details.
 
 After that, you'll need to grant EXECUTE privileges on functions explicitly:
 
 .. code-block:: postgres
 
+   GRANT EXECUTE ON ALL FUNCTIONS IN SCHEMA api TO web_user;
    GRANT EXECUTE ON FUNCTION login TO anonymous;
-   GRANT EXECUTE ON FUNCTION reset_password TO web_user;
```
steve-chavez, line 43: Could also be like
```sql
GRANT EXECUTE ON FUNCTION login TO anonymous;
GRANT EXECUTE ON FUNCTION signup TO anonymous;

-- you can also grant execute on all functions in a schema to a higher privileged role
GRANT EXECUTE ON ALL FUNCTIONS IN SCHEMA api TO web_user;
```

steve-chavez:
@tad-lispy Check my above comments :). After that I think the PR is good to merge.

tad-lispy:
Updated in the direction you suggested @steve-chavez.

steve-chavez:
Great work here @tad-lispy! Change already live http://postgrest.org/en/v7.0.0/schema_structure.html#functions.

tad-lispy:
Awesome! Thanks.

PR #339, opened 2020-07-11
Improve docs

wolfgangwalther:
Just a couple of changes related to:

https://github.com/PostgREST/postgrest/pull/1475

https://github.com/PostgREST/postgrest/pull/1552

https://github.com/PostgREST/postgrest/pull/1552#issuecomment-656710895


api.rst
```diff
@@ -1227,6 +1227,28 @@ as in ``{1,2,3,4}``. Note that the curly brackets have to be urlencoded(``{`` is
 
    In these versions we recommend using function parameters of type json to accept arrays from the client.
 
+Calling variadic functions
```
wolfgangwalther, line 1230: Not sure whether that whole thing actually needs it's own block or whether it can just be a note to the array part above?

api.rst
```diff
@@ -1148,13 +1148,13 @@ PostgreSQL has four procedural languages that are part of the core distribution:
 Immutable and stable functions
```
wolfgangwalther, line 1148: I tried to rephrase that a bit to make it more clear what is actually happening (the volatility marker has little real effect - except for POST actually), but I'm not sure I'm 100% happy yet. @steve-chavez what do you think?

api.rst
steve-chavez: Could be a note since it works the same way as the array case. Like: "You can also call a function with a variadic parameter in this way. Example... ".

Though the downside of a note is that it doesn't generate a hash identifier for the URL.

So I think subheading is good. You could link this subheading on the `releases/upcoming.rst`(on the entry you added).

api.rst
steve-chavez: The GET case looks better. However POST doesn't look that clear(gets a bit lost).

I think POST deserves it's own paragraph. Could be like: "For POST pgrst executes the statement in a write(read/write) transaction except for functions marked as immutable/stable...".

releases/upcoming.rst
```diff
@@ -18,6 +18,8 @@ Fixed
 
 * Fix showing UNKNOWN on ``postgrest --help`` invocation.
   |br| -- `@monacoremo <https://github.com/monacoremo>`_
+* Fix :ref:`_s_procs_variadic`.
```
wolfgangwalther, line 21: I don't have Sphinx installed, so I can't test that - is that the correct format for the link?

api.rst
wolfgangwalther: I think I got it now. What do you think?

steve-chavez: It's clearer. I like it :+1:

api.rst
```diff
@@ -1148,13 +1148,15 @@ PostgreSQL has four procedural languages that are part of the core distribution:
...
+PostgREST executes POST requests in a read/write transaction except for functions marked as ``IMMUTABLE`` or ``STABLE``. Those must not modify the database and are executed in a read-only transaction compatible for read-replicas.
+
+Procedures that do not modify the database can be called with the HTTP GET verb as well, if desired. PostgREST executes all GET requests in a read-only transaction. Modifying the database inside read-only transactions is not possible and calling volatile functions with GET will fail.
 
 .. note::
 
-  The volatility marker is a promise about the behavior of the function.  PostgreSQL will let you mark a function that modifies the database as ``immutable/stable`` without failure.  However the function will fail when called through PostgREST since it executes it in a read-only transaction.
+  The `volatility marker<https://www.postgresql.org/docs/current/static/xfunc-volatility.html>`_ is a promise about the behavior of the function.  PostgreSQL will let you mark a function that modifies the database as ``IMMUTABLE`` or ``STABLE`` without failure.  However, because of the read-only transaction this would still fail with PostgREST.
```
steve-chavez, line 1157: The link needs a space to render properly:
```suggestion
  The `volatility marker <https://www.postgresql.org/docs/current/static/xfunc-volatility.html>`_ is a promise about the behavior of the function.  PostgreSQL will let you mark a function that modifies the database as ``IMMUTABLE`` or ``STABLE`` without failure.  However, because of the read-only transaction this would still fail with PostgREST.
```

releases/upcoming.rst
steve-chavez: Needs the first underscore removed. Sphinx has its quirks.
```suggestion
* Fix :ref:`s_procs_variadic`.
```

wolfgangwalther:
Added docs for PostgREST/postgrest#1560

api.rst
```diff
@@ -1227,6 +1229,30 @@ as in ``{1,2,3,4}``. Note that the curly brackets have to be urlencoded(``{`` is
 
    In these versions we recommend using function parameters of type json to accept arrays from the client.
 
+.. _s_procs_variadic:
```
steve-chavez, line 1232: Hey @wolfgangwalther. Since we'll not include variadic in this release, could you leave this section for another PR so I can merge?

Recently, we had a user confused about volatility on [gitter chat](https://gitter.im/begriffs/postgrest?at=5f22fba3a050df430025ff2a). This addition to the docs would clear things up.

releases/upcoming.rst
```diff
@@ -18,6 +18,8 @@ Fixed
 
 * Fix showing UNKNOWN on ``postgrest --help`` invocation.
   |br| -- `@monacoremo <https://github.com/monacoremo>`_
+* Fix :ref:`s_procs_variadic`.
+  |br| -- `@wolfgangwalther <https://github.com/wolfgangwalther>`_
```
steve-chavez, line 22: Also here, for when variadic gets merged.

PR #345, opened 2020-08-19
Fix order of options in tar command

geoffreyvanwyk:
With the current order, the tar command thinks the name of the archive is J.

### DIFF
--- a/tutorials/tut0.rst
+++ b/tutorials/tut0.rst
@@ -46,7 +46,7 @@ The pre-built binaries for download are :code:`.tar.xz` compressed files (except
 
   # download from https://github.com/PostgREST/postgrest/releases/latest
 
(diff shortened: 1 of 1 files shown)


steve-chavez:
Oh, that one is strange. In Linux(NixOS) I get the correct name without changing the order:

```
tar xvfJ postgrest-v7.0.1-linux-x64-static.tar.xz
postgrest
```

@geoffreyvanwyk What's your OS and tar version?(Check with `tar --version`)

(Not opposed to merging the change, just curious)

geoffreyvanwyk:
Hi @steve-chavez I am on Windows 10 WSL2. My tar version is 1.29.

steve-chavez:
@geoffreyvanwyk Ah, I see. The `xJf` form is better for cross-compatibility then. I'll keep it in mind.

Thank you!

PR #346, opened 2020-08-19
Update `localhost` with docker's IP address.

viiicky:
I tried setting the local setup. I was building from source. I had to replace `localhost` with docker-machine's IP address to get it working.

### DIFF
--- a/tutorials/tut0.rst
+++ b/tutorials/tut0.rst
@@ -35,6 +35,12 @@ If Docker is not installed, you can get it `here <https://www.docker.com/communi
 
 This will run the Docker instance as a daemon and expose port 5433 to the host system so that it looks like an ordinary PostgreSQL server to the rest of the system.
 
+Make note of docker machine's IP:
+
+.. code-block:: bash
+
+  sudo docker-machine ip
+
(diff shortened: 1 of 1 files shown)


tutorials/tut0.rst
```diff
@@ -35,6 +35,12 @@ If Docker is not installed, you can get it `here <https://www.docker.com/communi
 
 This will run the Docker instance as a daemon and expose port 5433 to the host system so that it looks like an ordinary PostgreSQL server to the rest of the system.
 
+Make note of docker machine's IP:
+
+.. code-block:: bash
+
+  sudo docker-machine ip
+
```
steve-chavez, line 43: Hey @viiicky. What's your OS?

Personally I don't use `docker-machine` and I'm not sure when it's needed.

That being said, it may be worth to put this on a note "If you use docker-machine, you can get the IP with...".

This note could be closer to the Step 5: Run PostgREST.

viiicky: Hey, I realised lately, that these changes are specific to my dev setup. We need not put this up. Taking the PR down. Thanks.

Issue #340, opened 2020-07-16
Move the CHANGELOG to docs


steve-chavez:
Right now we duplicate the CHANGELOG from https://github.com/PostgREST/postgrest/blob/master/CHANGELOG.md to the docs. That's not ideal and gets cumbersome.

Instead we can have the CHANGELOG on the docs and have a link on the github release page.

I've seen godot does it like this: https://github.com/godotengine/godot/releases.

I believe this would be better since sometimes we need to link the CHANGELOG entry to the newly added doc section.

We can announce this change on the upcoming release.

(Got this from the discussion on https://github.com/monacoremo/postgrest-sessions-example/issues/19#issuecomment-639770101)

steve-chavez:
Another idea would be to use a convention on the git commits, and create a changelog based on that. Like https://github.com/semantic-release/semantic-release#commit-message-format.

wolfgangwalther:
Yes, please. But how are we going to handle references to docs sections with that?

wolfgangwalther:
Also looking at #367, where a lot of the new config options are missing: Have you considered moving the docs into the main repo?

This would allow us to:
- Make sure that every PR is merged with the corresponding docs
- Have the changelog in one repo only and also track those immediately with the PR
- Maybe even have a CI check to notify us if a Changelog entry is missing

This would make it a lot easier to track docs updates and make sure the latest docs are always up2date. This is even more important now, that we have nightly builds.

steve-chavez:
> Have you considered moving the docs into the main repo?

@wolfgangwalther It used to be that way, but it was split a long time ago.

The main problem is that the docs don't only contain PostgREST features. It also contains tutorials, how-tos, explanations, blog posts, links, landing page, etc. Having all the previous changes on a single repo would:

- Trigger a CI rebuild on a each change. e.g. a simple "typo fix" will do that. And we have several CIs(maybe one more later). It looks hard to control.
- Makes history on the main repo harder to track.

Btw, Remo also asked me this before: https://github.com/monacoremo/postgrest-sessions-example/issues/19#issuecomment-634897040.

steve-chavez:
Just noted that Godot keeps more than one CHANGELOG:

- Code changelog: https://github.com/godotengine/godot/blob/3.3/CHANGELOG.md
- Docs changelog: https://godotengine.org/article/release-candidate-godot-3-3-3-rc-1
- These seem autogenerated:
[...]

PR #365, opened 2020-11-22
Add note about embedding of view-chains and the interaction with `db-extra-search-path`

wolfgangwalther:
Following up on https://github.com/PostgREST/postgrest/pull/1625#issuecomment-728357780.

What do you think @steve-chavez?

### DIFF
--- a/api.rst
+++ b/api.rst
@@ -670,7 +670,7 @@ As an example, let's create a view called ``nominations_view`` based on the *nom
   FROM
     nominations;
 
-Since it contains ``competition_id`` and ``film_id``—and each one has a **foreign key** defined in its source table—we can embed *competitions* and *films*:
+Since it contains ``competition_id`` and ``film_id`` — and each one has a **foreign key** defined in its source table — we can embed *competitions* and *films*:
 
 .. code-block:: http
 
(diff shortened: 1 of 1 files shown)


api.rst
```diff
@@ -678,6 +678,8 @@ Since it contains ``competition_id`` and ``film_id``—and each one has a **fore
 
 It's also possible to embed `Materialized Views <https://www.postgresql.org/docs/11/rules-materializedviews.html>`_.
 
+Views can also depend on other views, which in turn depend on the actual source table. For PostgREST to pick up those chains recursively to any depth, all the views must be in the search path, so either in the exposed schema (:ref:`db-schema`) or in one of the schemas set in :ref:`db-extra-search-path`. This does not apply to the source table, which can be in any hidden schema as well.
```
steve-chavez, line 681: "hidden schema" is a bit mysterious. Maybe we could use "private schema" and link to http://postgrest.org/en/v7.0.0/schema_structure.html#schema-isolation?

steve-chavez: A subheading would also be nice, that way we can link to the new feature.

wolfgangwalther:
What about this? Better that way?

api.rst
```diff
@@ -670,7 +670,7 @@ As an example, let's create a view called ``nominations_view`` based on the *nom
   FROM
     nominations;
 
-Since it contains ``competition_id`` and ``film_id``—and each one has a **foreign key** defined in its source table—we can embed *competitions* and *films*:
+Since it contains ``competition_id`` and ``film_id`` — and each one has a **foreign key** defined in its source table — we can embed *competitions* and *films*:
```
wolfgangwalther, line 673: The main reason for this change was, because it broke my vim-syntax highlighting for the remaining file.. but I think it's better to read as well.

wolfgangwalther:
I did not bother to add something to `releases/upcoming.rst`, yet, given that most of the changelog is still missing. Tell me again, how were we going to proceed with this? You had something better in mind, didn't you?

steve-chavez:
@wolfgangwalther Main idea was to move the CHANGELOG to docs: https://github.com/PostgREST/postgrest-docs/issues/340. But still not sure about the implications..

PR #367, opened 2020-11-22
Add documentation for missing config options

wolfgangwalther:
Wanted to add documentation for `tx-rollback-all` and `tx-allow-override` and realized there were quite a few config options missing from the docs still.

Added placeholders for now, we gotta fill those. Suggestions welcome.

I also cleaned up the `postgrest --help` command a bit, while putting both the help and the docs next to each other. Consistent order in both now.

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -39,6 +39,8 @@ db-anon-role             String             Y
 db-pool                  Int     10
 db-pool-timeout          Int     10
 db-extra-search-path     String  public
+db-channel               String  pgrst
+db-channel-enabled       Bool    False
 server-host              String  !4
 server-port              Int     3000
 server-unix-socket       String
@@ -47,11 +49,15 @@ openapi-server-proxy-uri String
 jwt-secret               String
 jwt-aud                  String
 secret-is-base64         Boolean False
+role-claim-key           String  .role
 max-rows                 Int     ∞
 pre-request              String
-app.settings.*           String
-role-claim-key           String  .role
+root-spec                String
 raw-media-types          String
+log-level                String  error
+tx-rollback-all          Boolean False
+tx-allow-override        Boolean False
+app.settings.*           String
 ======================== ======= =========  ========
 
 .. _db-uri:
@@ -130,6 +136,16 @@ db-extra-search-path
   This parameter was meant to make it easier to use **PostgreSQL extensions** (like PostGIS) that are outside of the :ref:`db-schema`.
 
   Multiple schemas can be added in a comma-separated string, e.g. ``public, extensions``.
+
+.. _db-channel:
+
+db-channel
+----------
+
+.. _db-channel-enabled:
+
+db-channel-enabled
+------------------
 
 .. _server-host:
 
@@ -219,27 +235,6 @@ secret-is-base64
 
   When this is set to :code:`true`, the value derived from :code:`jwt-secret` will be treated as a base64 encoded secret.
 
-.. _max-rows:
-
-max-rows
---------
-
-  A hard limit to the number of rows PostgREST will fetch from a view, table, or stored procedure. Limits payload size for accidental or malicious requests.
-
-.. _pre-request:
-
-pre-request
------------
-
-  A schema-qualified stored procedure name to call right after switching roles for a client request. This provides an opportunity to modify SQL variables or raise an exception to prevent the request from completing.
-
-.. _app.settings.*:
-
-app.settings.*
---------------
-
-  Arbitrary settings that can be used to pass in secret keys directly as strings, or via OS environment variables. For instance: :code:`app.settings.jwt_secret = "$(MYAPP_JWT_SECRET)"` will take :code:`MYAPP_JWT_SECRET` from the environment and make it available to postgresql functions as :code:`current_setting('app.settings.jwt_secret')`.
-
 .. _role-claim-key:
 
 role-claim-key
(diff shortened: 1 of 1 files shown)


steve-chavez:
> realized there were quite a few config options missing from the docs still.

Yeah. I didn't get the chance of including them in the docs. My usual workflow was: when close to a release, review all the items on the CHANGELOG and start working on docs for them.

steve-chavez:
>  My usual workflow was: when close to a release, review all the items on the CHANGELOG and start working on docs for them.

You know, it's not DRY, but that was the best workflow for me. Keeping the main repo light on details on the CHANGELOG and eventually work on docs. Switching repos is maybe what's annoying.

Sometimes the features require a long explanation that would be big for the git commit as well. Like this one http://postgrest.org/en/v7.0.0/api.html#estimated-count.

wolfgangwalther:
Absolutely, makes a lot of sense! I kind of realized that a bit later as well, when I went through the issue list in the repo here. It's not that they were forgotten, the docs were just not written yet. And adding them to the list, like I did, with maybe a bit of text will still not solve it 100%, because some of those features need a lot more text in other places as well, as you say.

I wonder whether that workflow might need to be adjusted now that we have the nightly builds. It seems important that the latest docs are kind-of-in-sync with the nightly, what do you think?

steve-chavez:
Yeah, so we can agree that the docs is in fact another project with its own set of issues(ideas, enhancements, etc).

> I wonder whether that workflow might need to be adjusted now that we have the nightly builds. It seems important that the latest docs are kind-of-in-sync with the nightly, what do you think?

Hmm.. maybe the nightly builds docs would be fine with just the main repo CHANGELOG(the issues are also there for context)? It's likely that users wanting to use the nightly have been following the issue.

That way we don't have to worry about switching to the docs whenever we do a feature on the main repo.

(In this case, the doc updating entry from the PR template should be removed as well: https://raw.githubusercontent.com/PostgREST/postgrest/master/.github/PULL_REQUEST_TEMPLATE.md)

Might be good enough. I don't have a better workflow for now.

wolfgangwalther:
> Yeah, so we can agree that the docs is in fact another project with its own set of issues(ideas, enhancements, etc).

Yes, absolutely. Merging repos (as suggested in the changelog issue) is not an option, I see the value of separate repos. Far more positives.

> Hmm.. maybe the nightly builds docs would be fine with just the main repo CHANGELOG(the issues are also there for context)? It's likely that users wanting to use the nightly have been following the issue.
> 
> That way we don't have to worry about switching to the docs whenever we do a feature on the main repo.
> 
> (In this case, the doc updating entry from the PR template should be removed as well: https://raw.githubusercontent.com/PostgREST/postgrest/master/.github/PULL_REQUEST_TEMPLATE.md)
> 
> Might be good enough. I don't have a better workflow for now.

For me, the workflow of implementing features and delaying docs until the next release does not work for two reasons:

- Delaying the docs update is essentially pushing the responsibility away from the committer to the maintainer. I do now have a bit of responsibility as well, but I don't want to give you a huge list of things to write docs for, before you can release something.

- Now that we have nightly builds and a bit of a backlog on the docs updates - every feature we push to the main repo will essentially make it harder to get a release together, as there is more and more to do for it. And the pressure to do so is less as well, because we have the nightly.

It's highly unlikely that any docs will be written for a feature before it is implemented (although that could be an interesting workflow... :) ), so by nature the docs repo will always lag behind. We should strive to try to reduce the lag. Personally, that means I will try to get those docs PRs together asap after merge of the main repo PR, but I am absolutely fine with a different approach for you.

I suggest the following, though:

- Keep the doc updating entry in the PR template, maybe even turn it into an automated TODO checklist. I will open a PR with a suggestion for this.

- Replace the main repo CHANGELOG with good commit messages as suggested here: https://github.com/PostgREST/postgrest-docs/issues/340#issuecomment-731863705. This will reduce friction (merge conflicts) in the main repo as well, because merging multiple smaller fixes is a pain right now with the changelog entry conflicting very often.

- Write the feature docs without the `release/upcoming` stuff, but create a script to parse those commit messages into a template for a new release notes file. This can then be adjusted for docs references. This script can be run when a new version is released. In the docs, the content of the upcoming release notes would point to the github release page for the nightly release.

- Add a similar script to the nightly ci step to update the nightly release page with the list of commits since the last official release. So the nightly release page would always have the latest changes and we link to it.

This way we only have to keep one list of changes up2date - via commit messages. That should be easy enough. And the latest docs will always point to the latest changes, even if we don't update them in a timely manner, because they are pointing to the github release page. If we add a note to the upcoming page, that the docs might not be in sync with those changes yet, that should be good enough.

This should allow us both to go about the docs mostly in our own way and simplify a couple of things.

What do you think?

wolfgangwalther:
Actually the nightly release could maybe be update with this:
https://probot.github.io/apps/release-drafter/

This seems like we should be able to set it up for our needs?

wolfgangwalther:
If we opted for conventional commits, we could use https://probot.github.io/apps/semantic-pull-requests/ as well. Of course could also just roll our own, by just adding 3 prefixes `added:`, `changed:`, `fixed:` that correspond to the categories in our current changelog and ignore all other commits, that don't have those prefixes.

Maybe we can also open up a 4th group `perf:`, to make **the performance release** more visible :o)

steve-chavez:
I like the idea of treating the commit messages as the source of truth :). Also agree about not touching the `upcoming` page and leaving it with a reference to the nightly release.

I haven't used any of the github bots. But I like the direction in general.

If it's hard to automate everything at first, we can also leave certain steps to be done manually.

> but create a script to parse those commit messages into a template for a new release notes file

It'd be great to have this one in Nix. Similarly to `postgrest-release-dockerhubdescription`, which can be tried manually and works every time.

wolfgangwalther:
just a rebase, no changes, to see what's going on with CI.

wolfgangwalther:
Works fine here and makes me aware of my shortcomings in underlining titles correctly. Would also need to add to the `postgrest.dict` here.

wolfgangwalther:
I do not intend to work on or merge this PR, but will keep it open for now, because of the discussion around our changelog handling. Creating the tooling for that, is still on my todo-list.

The actual docs for config options will be written in different PRs.

PR #372, opened 2020-11-29
Add nix-shell tools, fix circleci, resolves #322

### LINKED ISSUE #322: Fix CircleCI
Circle has been failing for a while here, I think it has to do with the new version migration. It should be fixed to get `aspell` running again.

wolfgangwalther:
Getting my feet wet with nix a bit.

@monacoremo would love to have feedback on the nix stuff.


wolfgangwalther:
@steve-chavez I have enabled pipelines for PRs in circleci.

wolfgangwalther:
@monacoremo so it's failing in CI because git can't be found. The checkout step worked, so some kind of git was available.

I guess this is because of `with import (builtins.fetchGit {` in `default.nix`. That was in there before already, and I am not sure whether I need it at all.

I guess I could fix it quite easily by changing the docker image in ci from nixos/nix to ubuntu or something. However, this would not help my learning. I copied all the parts together from stuff in the postgrest repo, that you did, so I don't really have a clue whats happening. What I am specifically confused about is:
- CI uses a docker nixos/nix image (copied from `style-check` in postgrest CI)
- `default.nix` then downloads another nixos environment in a different version

This seems like I am running nix inside nix - can this not be done simpler?

I also noticed that the CI setup in `postgrest` has some of the nix steps with a nixos/nix base image and others with a "machine: true" (so ubuntu) where nix is installed manually. I wondered whether there was any specific reason for that?

Please enlighten me!

monacoremo:
Many good questions! I think it's going to be great to learn, almost there :-) 

I'll answer a few questions here, and also comment on some of the changes:

> I guess this is because of with import (builtins.fetchGit { in default.nix. That was in there before already, and I am not sure whether I need it at all.

Not quite sure what is happening here, but yes, it's probably because of the fetchGit and we don't need it like that. More on that on the next question.

> This seems like I am running nix inside nix - can this not be done simpler?

There are two pieces that we need:

* The Nix tooling that includes `nix-build`, `nix-shell`, the `/nix/store` etc. This we get either through a Docker image or by using the Nix install script
* A version of https://github.com/NixOS/nixpkgs . One already comes with the Nix tooling usually, but we can't be sure which one it is exactly. To be absolutely sure, we "pin" our own version of Nixpkgs, as identified by a commit hash. This is what the `fetchGit` part above is trying to do. In Postgrest, the same is happening here: https://github.com/PostgREST/postgrest/blob/9adec12a678afceca78f6bfebba236b031e63487/default.nix#L17-L42

Essentially what happens there is:

* We download a specific Nixpks tarball, which is more efficient than cloning the git repository, from e.g. https://github.com/nixos/nixpkgs/archive/3a477cf32417b4c81c9fe42fbfb52c40b69fc110.tar.gz
* We check that the hash of the downloaded tarball is what we expect (making absolutely sure it's the same version)
* The tarball is unpacked
* We define overlays that we want to use to reshape Nixpkgs to our need (not relevant here for postgrest-docs, skip)
* We import our fresh Nixpkgs version and use the packages defined there as dependencies

You can try these steps manually also, running `nix-build -f your/unpacked/nixpkgs -A pkgs.hello` in the end to use your new Nixpkgs version.

This "pinning" is needed to get reproducibility, and absolutely worth it IMO!

> I also noticed that the CI setup in postgrest has some of the nix steps with a nixos/nix base image and others with a "machine: true" (so ubuntu) where nix is installed manually. I wondered whether there was any specific reason for that?

I used `machine: true` there simply because it runs with a non-root user by default (required for running our temporary PostgreSQL dbs) and CircleCI is a bit more generous with RAM or CPUs if I remember correctly. Here the lighter Nixos image should be fine.

default.nix
```diff
@@ -4,14 +4,45 @@ with import (builtins.fetchGit {
   rev = "95fed28ac372c61eb83c87ad97c24b0f957827bf";
 }) {};
 
+let
+  build =
+    writeShellScriptBin "postgrest-docs-build"
+      ''
+        set -euo pipefail
+
+        sphinx-build -b html -a -n . _build
```
monacoremo, line 13: This is not going to work outside nix-shell (e.g. when using `nix-env -iA`), as sphinx-build is not taken from nixpkgs. You need to do something like `${pythonPackages.sphinx}/bin/sphinx-build ...`

wolfgangwalther: Ah, thanks for the explanation. I was kind of confused why it **did** work, because in your scripts you always prefix that, and it kind of made sense... - but I only tried inside nix-shell.

wolfgangwalther:
> * The Nix tooling that includes `nix-build`, `nix-shell`, the `/nix/store` etc. This we get either through a Docker image or by using the Nix install script
> 
> * A version of https://github.com/NixOS/nixpkgs . One already comes with the Nix tooling usually, but we can't be sure which one it is exactly. To be absolutely sure, we "pin" our own version of Nixpkgs, as identified by a commit hash. This is what the `fetchGit` part above is trying to do. In Postgrest, the same is happening here: https://github.com/PostgREST/postgrest/blob/9adec12a678afceca78f6bfebba236b031e63487/default.nix#L17-L42

Ah, that clear separation between nix tooling and nixpkgs is what I didn't really see. So it doesn't matter that much which version of nix-tooling we use (so we can either use a base docker image or pull the nix env like you do in postgrest ci), but we need to pin the nixpkgs. Ok.

How would I go about it in this case? Do I just add the same variable nixpkgs, that you have referenced here, and then add that to buildInputs? Or do I need to somehow prefix the `python36Full` etc, with `nixpkgs.` or something? Not sure how those `buildInputs` really work.

> I used `machine: true` there simply because it runs with a non-root user by default (required for running our temporary PostgreSQL dbs) and CircleCI is a bit more generous with RAM or CPUs if I remember correctly. Here the lighter Nixos image should be fine.

Ah yes, of course. That makes a lot of sense, I had the same thing with the docker image for nix on windows. Postgres won't run with root.

default.nix
```diff
@@ -4,14 +4,45 @@ with import (builtins.fetchGit {
...
+        python livereload_docs.py
+      '';
+
+
+  spellcheck =
+    writeShellScriptBin "postgrest-docs-spellcheck"
+      ''
+        set -euo pipefail
+
+        cat *.rst | grep -v '^\(\.\.\|  \)' | sed 's/`.*`//g' | aspell -d ${aspellDicts.en}/lib/aspell/en_US -p ./postgrest.dict list | tee misspellings
```
wolfgangwalther, line 31: In theory I would need to prefix `cat`, `grep` etc. as well, right? Am I going to do that - or where do I draw the cut?

wolfgangwalther:
> How would I go about it in this case? Do I just add the same variable nixpkgs, that you have referenced here, and then add that to buildInputs? Or do I need to somehow prefix the `python36Full` etc, with `nixpkgs.` or something? Not sure how those `buildInputs` really work.

Ah I found a way that works. Does it make sense what I did?

wolfgangwalther:
But I get a new error and I have a feeling that the whole `stdenv.mkDerivation` is not really useful for `nix-env -iA`. It works with `nix-shell`, but that's about it, I guess. I will try a few things.

default.nix
monacoremo: A few standard utilities are on the path by default, see https://nixos.org/manual/nixpkgs/stable/#sec-tools-of-stdenv . You can easily test whether you got it right by using `nix-shell --pure ...`

monacoremo:
> Ah I found a way that works. Does it make sense what I did?

Looks good generally! You can simplify a bit and as a matter of taste I would not use `with`. You could do:

```
let
  nixpkgsVersion = 
    ...

  pkgs =
    import (fetchTarball ...) { };
```

without a second `let ... in ...`. Then you will need to prefix each package used with `pkgs.`, which I find less confusing then pulling all pkgs into scope with `with`. (later we could have a look at defining modules with callPackage, where you avoid prefixing again, but we will not need it here I think)

.circleci/config.yml
```diff
@@ -0,0 +1,33 @@
...
+
+jobs:
+  build:
+    docker:
+      - image: nixos/nix:2.3
+    steps:
+      - checkout
+      - run:
+          name: Install build script
+          command: nix-env -f default.nix -iA build
```
monacoremo, line 11: This requires that our nix expression in default.nix has an attribute `build`, which it currently cannot have I believe.

We have two options:
* Split the derivation at the end, which is an environment intended purely for nix-shell, into its own `shell.nix` file (see postgrest for this pattern). Then we fix default.nix to return a set at the end: `... in { inherit build spellcheck; }` (which is the same as `{ build = build; spellcheck = spellcheck; }`.
* We can be a bit sneaky and just set these attributes on the derivation we are already returning. We would notice at Nix eval time if that broke anything. To do that, we can just append `// { inherit build spellcheck; }` to default.nix ("merge the set before with this new set that has build and spellcheck as attributes, with the right hand side overriding the left hand side if needed").

wolfgangwalther: I think I did something inbetween. What do you think?

(Only reading your comment right now after pushing the latest changes)

wolfgangwalther:
> Looks good generally! You can simplify a bit and as a matter of taste I would not use `with`. 

Ok, will do that.

default.nix
```diff
@@ -1,17 +1,72 @@
...
+         | grep -v '^\(\.\.\|  \)'                                                             \
+         | sed 's/`.*`//g'                                                                     \
+         | ${aspell}/bin/aspell -d ${aspellDicts.en}/lib/aspell/en_US -p ./postgrest.dict list \
+         | tee misspellings
+        test ! -s misspellings
+      '';
+in
+
+{
+  build = { inherit build; };
```
monacoremo, line 59: I think `inherit build serve spellcheck;` instead of the three lined here should to the trick!

wolfgangwalther: Unfortunately no, then `nix-shell` will complain about too many derivations. I put all 3 in a subkey `tools` at first, but this would only allow me to install all 3 of them together in CI. That's why I split them up.

wolfgangwalther: The error is: `error: nix-shell requires a single derivation`

wolfgangwalther: But your sneaky suggestion from above works. and looks better.

default.nix
wolfgangwalther: Ah `--pure` is nice. That helps a lot. Works!

wolfgangwalther:
So this seems to work now. The failing build is for real (because of spellcheck and build warnings).

Anything else you notice in the .nix file @monacoremo, or is it fine like this?

I might try adding a cache for /nix in circleci, because the "build" step feels like it's taking way too long to install dependencies. Spellcheck is almost instant.

wolfgangwalther:
> I might try adding a cache for /nix in circleci, because the "build" step feels like it's taking way too long to install dependencies. Spellcheck is almost instant.

@monacoremo Or is it better to go the cachix way here? In theory we could push those dependencies to the `postgrest` cachix cache, right?

monacoremo:
> I might try adding a cache for /nix in circleci, because the "build" step feels like it's taking way too long to install dependencies. Spellcheck is almost instant.

As we use pretty normal packages here, i would expect everything we need to be in the standard nixpkgs cache. Could you try:
* Using a more visible Nixpkgs version, e.g. 95c4f23725ba1158b24644cbf2406d113de1ef7e from https://github.com/NixOS/nixpkgs/tree/release-20.09 (poor man's solution to find the new hash: plug in any new value of correct length by changing the current hash, see what the nix-shell error message tells you it expects; proper way see nixpkgs-upgrade in postgrest)
* Use the default Python (3), instead of the old python36 (`python3` or maybe even just `python` should work?)

default.nix
```diff
@@ -1,17 +1,64 @@
...
+    };
+
+  # Nix files that describe the Nixpkgs repository. We evaluate the expression
+  # using `import` below.
+  pkgs = import (fetchTarball {
+      url = "https://github.com/nixos/nixpkgs/archive/${nixpkgsVersion.rev}.tar.gz";
+      sha256 = nixpkgsVersion.tarballHash;
+    }) {};
+
+  python = pkgs.python36.withPackages (ps: [ps.sphinx ps.sphinx_rtd_theme ps.livereload]);
```
monacoremo, line 16: Good job, not too obvious and easy to figure that out :-)

wolfgangwalther: I wonder whether that creates the caching situation? Does that not effectively create a python derivation with exactly those packages? So to have caching for that, somebody else would have to do the exact same thing...?

I will try your suggestions from above of course.

default.nix
```diff
@@ -1,17 +1,64 @@
...
+        cat *.rst                                                                              \
+         | grep -v '^\(\.\.\|  \)'                                                             \
+         | sed 's/`.*`//g'                                                                     \
+         | ${pkgs.aspell}/bin/aspell -d ${pkgs.aspellDicts.en}/lib/aspell/en_US -p ./postgrest.dict list \
+         | tee misspellings
+        test ! -s misspellings
+      '';
+in
+
+pkgs.stdenv.mkDerivation {
```
monacoremo, line 54: Nitpick: We could use pkgs.buildEnv here, making it clearer that we just want to put together an environment, e.g.

```
buildEnv {
  name = "postgrest-docs";
  paths = [ build serve spellcheck ];
}
```

default.nix
```diff
@@ -1,17 +1,64 @@
...
+
+        ${python}/bin/python livereload_docs.py
+      '';
+
+  spellcheck =
+    pkgs.writeShellScriptBin "postgrest-docs-spellcheck"
+      ''
+        set -euo pipefail
+
+        cat *.rst                                                                              \
```
monacoremo, line 45: alignment on the '\' got messed up already, why not just with one space?

monacoremo:
> Anything else you notice in the .nix file @monacoremo, or is it fine like this?

I like it, great entrance into the Nix world! :-)

Maybe one pass of nixpkgs-fmt to autoformat consistently to postgrest?

default.nix
monacoremo: my bet is on the 3.6 Python version

wolfgangwalther:
> Using a more visible Nixpkgs version, e.g. 95c4f23725ba1158b24644cbf2406d113de1ef7e from https://github.com/NixOS/nixpkgs/tree/release-20.09 (poor man's solution to find the new hash: plug in any new value of correct length by changing the current hash, see what the nix-shell error message tells you it expects; proper way see nixpkgs-upgrade in postgrest)

Strange. This does not do anything. I put in a wrong hash, start nix-shell and it not even rebuilds. No error whatsoever.

monacoremo:
> Strange. This does not do anything. I put in a wrong hash, start nix-shell and it not even rebuilds. No error whatsoever.

Weird. You did change the rev also?

wolfgangwalther:
Just edited my comment. I only changed the rev.

default.nix
wolfgangwalther: Hm. If I do that, then I can't use `inherit` anymore and the tools are not available via `nix-env -iA build` etc.

monacoremo:
> Just edited my comment. I only changed the rev.

Need to also change the hash arbitrarily, the download of the tarball is cached based on it.

default.nix
wolfgangwalther: You seem to know your craft... newer python worked :)

monacoremo:
Perfect, everything got pulled from the Nixpkgs cache!

wolfgangwalther:
Yes. Now only the `mkDerivation` vs `buildEnv` is open. Not sure about that.

monacoremo:
> Yes. Now only the `mkDerivation` vs `buildEnv` is open. Not sure about that.

Lets leave it for now, more cosmetic than an actual issue 

With the `// { ... }` approach it would work I think

wolfgangwalther:
Ah, changed it to `buildEnv` and now I'm installing just with `nix-env -f ... -i` and not `-iA build` etc.

This does pull in a few more dependencies for spellcheck, but that shouldn't be a problem. I originally wanted to avoid the live_reload deps to load in CI, but that ship has sailed with the `python.withPackages` anyway.



> With the `// { ... }` approach it would work I think

Ah, I have to look that up. When I first read it, I thought this was some kind of comment, but // does probably mean something else...

wolfgangwalther:
Of course `//` is just what I was thinking about earlier.. "how could I...?". Good to know, but I like what we have right now.

wolfgangwalther:
Ah really cool - that was fun. I will fix the errors that CI is now reporting, later and I guess then this can be merged. Thanks @monacoremo for the help - I learned a lot today.

wolfgangwalther:
Eh, something is still wrong. When I do `nix-shell --pure`, none of the 3 `postgrest-docs-` commands are available. Without pure they are, because I installed them with `nix-env -f ... -i` to test that.

Should the `paths` in buildEnv not take care of that? I'm confused.

wolfgangwalther:
I added a `shell.nix` file. Now, this looks to me like it should.

monacoremo:
That's exactly what I was going to suggest! Now we're fully on best practice and it should obviously work, even if it's one extra file

default.nix
```diff
@@ -1,17 +1,59 @@
...
+        FILES=$(find . -type f -iname '*.rst' | tr '\n' ' ')
+
+        cat $FILES \
+         | grep -v '^\(\.\.\|  \)' \
+         | sed 's/`.*`//g' \
+         | ${pkgs.aspell}/bin/aspell -d ${pkgs.aspellDicts.en}/lib/aspell/en_US -p ./postgrest.dict list \
+         | tee misspellings
+        test ! -s misspellings
+      '';
+in
```
monacoremo, line 56: To make it absolutely perfect, but a matter of taste: We can avoid the `inherit`ing so many values by doing

```
let
  ...

  python =
    ...;
in
{
  inherit pkgs;

  build =
    ...;

  ...
}
```

(if you ever need to refer to an item in the set from within the set, you will be able to do:

```
...
in
rec {
  ...
}
```
)

shell.nix
```diff
@@ -0,0 +1,16 @@
...
+  pkgs =
+    docs.pkgs;
+in
+pkgs.mkShell {
+  name = "postgrest-docs";
+
+  buildInputs = [
+    docs.build
+    docs.serve
+    docs.spellcheck
```
wolfgangwalther, line 14: Do you have any idea how to avoid repetition here? I feel like adding `docs.anotherTool` every time could be avoided?

In javascript terms, I would just like to do `[ ...docs ]`, or something like that. Although one of them is a set, the other one a list..... something like `buildInputs = valuesOf docs` maybe?

wolfgangwalther: Not sure what happens when I add `docs.pkgs` to buildInputs as well, though.

monacoremo: > Not sure what happens when I add docs.pkgs to buildInputs as well, though.

Could be fun, absolutely all of the 60k+ packages I think :-)

Otherwise you could use `builtins.attrValues` to turn the set from default.nix into a list.

I think it's good to be explicit with what we include in shell.nix. If we have many related scripts, we can group them with pkgs.buildEnv in default.nix.

wolfgangwalther: > Could be fun, absolutely all of the 60k+ packages I think :-)

I had a feeling there... :)

Okay, I stick with this then. Thanks!

wolfgangwalther:
Apart from learning a lot about nix, I did the following here:

- Added the `-W` flag on the build for the ci job, to turn warnings into errors. Then fixed those.

- Obviously fixed all the spelling stuff. Some of that was by keeping `postgrest.dict` up2date, I also added a small helper `postgrest-docs-dictcheck` that scans this dict for obsolete entries that are not used anymore. Found a couple and removed those.

I'm going to merge this now... and then we have CI back on the docs! Boom!

wolfgangwalther:
So the pipeline passed here, but is failing on master, because of some git / ssh error. Nice. Gotta investigate.

wolfgangwalther:
I have no clue whats going on. Re-ran the pipeline on master twice, failed both times. Worked in another PR just fine after rebasing.

PR #373, opened 2020-11-29
make docs full-width, resolves #351

### LINKED ISSUE #351: Wide style for docs
The docs don't look on wide monitors. See: https://github.com/readthedocs/sphinx_rtd_theme/issues/295

I've tried modifying the max-width before with no success.

The Linux Kernel does manage to widen the pages: https://www.kernel.org/doc/html/latest/filesystems/tmpfs.html.
We should try to get a similar style.

wolfgangwalther:
While this "works", I'm not 100% convinced, yet. We tend to not have as much text and those single sentences stretched out across the whole screen are not really better to read. Do we need to do more styling here?

I tried just copying in all the linux-docs css overrides (as this style as mentioned in #351), but I'm not sure of that either.

What do you think @steve-chavez?

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -1,3 +1,15 @@
+.wy-nav-content {
+  max-width: initial;
+}
+
+#postgrest-documentation {
+  max-width: 800px;
+}
+
+#postgrest-documentation > h1 {
+  display: none;
+}
+
 div.wy-menu.rst-pro {
   display: none !important;
 }
@@ -10,11 +22,11 @@ div.line-block {
   margin-bottom: 0px !important;
 }
 
-#sponsors{
+#sponsors {
   text-align: center;
 }
 
-#sponsors h1{
+#sponsors h2 {
   text-align: left;
 }
 
--- a/index.rst
+++ b/index.rst
@@ -1,5 +1,8 @@
 .. title:: PostgREST Documentation
 
+PostgREST Documentation
+=======================
+
 .. figure:: _static/logo.png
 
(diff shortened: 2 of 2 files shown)


steve-chavez:
Yeah, the wide style is not looking good on all the pages.

I think the main issue is the home page. I find the API page better in wide because I can see the operators description without scrolling.

Some screenshots:

#### Previous home(too much white space)

prior_home

#### New home(the logo is too big)

new_home

#### Previous API operators(needs scrolling, too much white space)

prior_api

#### New API operators:

new_api

steve-chavez:
Maybe we need centering and a bigger font instead. Not sure for now.

wolfgangwalther:
Yeah, it was the homepage that I thought didn't look that nice as well. Logo is just too big. The kernel doc feels like it's a bit bigger of a font-size, but it's actually not. They just use `serif` as family, but when I tried, I didn't like it.

> Maybe we need centering and a bigger font instead, though I'm not sure if it's a simple change.
> 
> Another option could be applying the wide style for all the pages besides the home page?

I tried centering and bigger fonts, but that seemed like the wrong direction. I went with the wide style for everything except the home page for now. I added a headline to the homepage to be able to target the css. We could also set the headline itself to `display: none`, although I don't really mind.

steve-chavez:
> I went with the wide style for everything except the home page for now.

Nice, that works for me. However I see the Sponsors header centered now:

home

> We could also set the headline itself to display: none, although I don't really mind.

The `display:none` would be nice. The home is also serving as a landing page for now.

wolfgangwalther:
Added display none and fixed the sponsor headline!

PR #374, opened 2020-11-29
Fix CI failing on master

wolfgangwalther:


wolfgangwalther:
So the problem is, that on master the SSH key for github is available. Once I enabled "Pass secrets to pull requests", the builds are failing here as well.

wolfgangwalther:
And now, after installing git and openssh in the nixos image, I get a different error:

```
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.
```

@steve-chavez any idea what we're using that SSH key that is stored in CircleCI for this repo, for? It does not seem to be to checkout code, because that works without the key. It seems it's not for deployment either. Is the key maybe just outdated?

steve-chavez:
>  Is the key maybe just outdated?

@wolfgangwalther It's possible, I never touched CI here(only saw it working on circle v1). 

> any idea what we're using that SSH key that is stored in CircleCI for this repo, for? 

At first I thought it could be related to https://readthedocs.org/projects/postgrest/, but no. I think it can be deleted. If it keeps failing maybe disable circle for the repo and then enable it again(worked for travis once).

wolfgangwalther:
So it seems like the key is needed. I deleted it and now it errors with `ssh: no key found`. :)

Unfortunately my privileges are not enough to disable circle for the repo, so I think you have to do that. Might have to add another deploy key as well ;)

wolfgangwalther:
@steve-chavez just for completeness, because you basically have to take over here:
The changes in this PR are not needed imho - those were just to test out a bit. Looking at the circle pipeline for the main repo, I'm pretty sure that the pipeline config as it is on master right now should work.

You can test success of resetting circle / deploy keys by triggering a rebuild on circle for master manually. If this goes beyond the checkout step without error, then you fixed the problem successfully.

Going to close this PR for now, but if for whatever reason those changes don't work for master, but do work for this PR here (you can retrigger manually as well), feel free to re-open.

Please don't forget to deactivate "pass secrets to pull requests" again, when everything is done, this is just enabled to reproduce it here.

steve-chavez:
@wolfgangwalther New deploy key was added and now checkout is working on master!

> Please don't forget to deactivate "pass secrets to pull requests" again

Done!

Issue #328, opened 2020-06-01
Autobuild docs on PRs


steve-chavez:
From https://docs.readthedocs.io/en/stable/guides/autobuild-docs-for-pull-requests.html.

This would be nice for contributions.

wolfgangwalther:
Works now!

PR #376, opened 2020-11-30
reference: Add response.status GUC

wolfgangwalther:
@steve-chavez I just went through all the branches in this repo and deleted some 4 years old feature branches that were already merged. Now it's only master and the various `vX.Y` branches remaining. Those version branches have some different commits, so I guess those should be kept around to still provide the old docs on readthedocs.

However we also have the `preview` branch. This is not linked on readthedocs, although it is available via URL. It has only 1 commit more than master (but is way behind in others). I think the change here is good, any reason why you committed that on `preview` and not `master`?

Can you maybe shed a bit more light on this branch in general? Is that still a thing or was that just an idea that is outdated by now?

### DIFF
--- a/api.rst
+++ b/api.rst
@@ -1496,7 +1496,7 @@ Notice that the variable should be set to an *array* of single-key objects rathe
 .. _pre_req_headers:
 
 Setting headers via pre-request
--------------------------------
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 By using a :ref:`pre-request` function, you can add headers to GET/POST/PATCH/PUT/DELETE responses.
 As an example, let's add some cache headers for all requests that come from an Internet Explorer(6 or 7) browser.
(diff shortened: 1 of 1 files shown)


wolfgangwalther:
(by the way, I kind of like the fact that I open up a PR with **your changes** in here for review. hahaha)

wolfgangwalther:
Now that we have CI, we would have to rebase this on master of course, before merging, to make sure your spelling was alright... ;)

steve-chavez:
:D. I used the preview branch to see how the docs looked like "live". It's hidden on readthedocs, but it does show here: http://postgrest.org/en/preview/

It's the lazy alternative to https://github.com/PostgREST/postgrest-docs/issues/328, I just force push on preview. Autobuild would be much better.

> Now that we have CI, we would have to rebase this on master of course

This commit should already be included on master. I see http://postgrest.org/en/latest/api.html#setting-response-status-code.

wolfgangwalther:
> Autobuild would be much better.

I had a quick read on that at some point the past days and it looked like a thing that just needs to be set up in the readthedocs settings. Is that true? Or do we need to do anything else about it? I don't have the privileges on readthedocs, but I would certainly like to make this happen - anywhere I can help?

> This commit should already be included on master. I see http://postgrest.org/en/latest/api.html#setting-response-status-code.

Everything is in fact there. I just don't get why this PR says it has no merge conflicts and does show a difference in "Files Changed". The difference should basically be 0? Very confusing...

steve-chavez:
It's cool to delete the `preview` branch though. I don't see myself using it anymore.

The local setup with Nix will work great.

wolfgangwalther:
> It's cool to delete the `preview` branch though. I don't see myself using it anymore.

Will have to remove that from readthedocs as well, right?

steve-chavez:
@wolfgangwalther Do you have an account on readthedocs? I can add you as a maintainer there.

wolfgangwalther:
Uh, eh... maybe you already did. Just logged in and now I see some "Administration" stuff. Didn't see that yesterday?

wolfgangwalther:
ah ok :D

wolfgangwalther:
While we're on it - could you change the "About" section on this repo? (I'm not an admin on github :D).

It currently looks like this:
```
About

Documentation for
github.com/begriffs/postgrest 
```

The link is redirected, but the current PostgREST link would look better!

PR #378, opened 2020-11-30
remove sections for manually creating the test database and building with stack-in-docker

wolfgangwalther:
https://github.com/PostgREST/postgrest/pull/1665 removes `test/destroy_test_db` and `test/Dockerfile.test`, so this removes the sections in the docs that reference it.

Of course this has to be replaced with documentation for our awesome nix dev environment (hi @monacoremo :D), but first I wanted to make sure that we're not pointing to tools that are not available anymore.


wolfgangwalther:
@monacoremo I guess you would be suited for the task to write some stuff about the nix dev env. Should we just point to the `nix/Readme.md` in the meantime? Your call.

monacoremo:
> Should we just point to the `nix/Readme.md` in the meantime?

Yes, let's link to the nix/readme for now! 

Not sure yet what the focus in the docs should be vs the readme. Happy to discuss if you have thoughts on that!

wolfgangwalther:
> > Should we just point to the `nix/Readme.md` in the meantime?
> 
> Yes, let's link to the nix/readme for now!
> 
> Not sure yet what the focus in the docs should be vs the readme. Happy to discuss if you have thoughts on that!

I thought more about that. I feel like we shouldn't replicate the nix docs in both places. I also think they need to be in the main repo, for a random contributor to come along and just be able to work with it, before hitting the docs for it. The readme is really good.

So... maybe we should just replace the whole `development.rst` section in the docs with a short `contributing.rst` that links to the nix-readme and mentions a couple of other resources (docs repo, commit guidelines and stuff). Then in the main repo we could add a small section `Contributing` to the main readme as well - probably the same content. And all the nix stuff stays right where it is put.

WDYT?

monacoremo:
Fully agreed - it should be in the main repo and we should not duplicate, so what you propose is the best way to go!

wolfgangwalther:
First step: I had a look at what was left over in `development.rst` after removing the obsolete parts here. While we don't need the two sections "Build from Source" and "Running the Test Suite" for our development env anymore, I thought it would still be nice to keep them - for those folks:

> When a pre-built binary does not exist for your system you can build the project from source.

I took those two sections and integrated them into the "Installation" chapter.

Next step will be to write that "Contributing" part.

wolfgangwalther:
Got a bit side-tracked here, because I noticed some links were outdated. So I added `postgrest-docs-linkcheck` which uses sphinx `linkcheck` to do exactly that. Did a major cleanup of all external links with that. The tool could be added to CI, but currently it's hitting rate limits at github.com, because we have so many of those links. Support for rate limiting is already included upstream, but unreleased - so maybe we can add that later to CI.

wolfgangwalther:
So, this should be good to merge right now. Could you have a look before, @steve-chavez?

Here's a summary what I did:
1. Removed `development.rst`, which was heavily outdated after PostgREST/postgrest#1665.
2. Moved a part of the development chapter ("Building from Source") to `install.rst` as another way of installing PostgREST.
3. Left all the "how to run the tests" out - this is replaced by the docs on our nix dev environment.
4. Added a "Contributing" part to `index.rst` that just links to our Contributing Guidelines on GitHub. From there we will need to link to the nix Readme (right now, we're linking back to the removed docs page).
5. Bonus: Added `postgrest-docs-linkcheck`, ran it and fixed all kinds of outdated links. Reformatted a bit here and there.

Steve, I'm mostly looking for feedback on 2-4. No need to go through all the changed links ;)

The goal was to reduce duplication of content going forward, that's why the docs now only link to the main repo for everything related to the development/contributing.

ecosystem.rst
```diff
@@ -19,27 +19,26 @@ Community Tutorials
...
-* `myfreeweb/moneylog <https://github.com/myfreeweb/moneylog>`_ - accounting web app in Polymer + PostgREST
-* `tyrchen/goodfilm <https://github.com/tyrchen/goodfilm>`_ - example film api
-* `begriffs/postgrest-example <https://github.com/begriffs/postgrest-example>`_ - sqitch versioning for API
-* `SMRxT/postgrest-demo <https://github.com/SMRxT/postgrest-demo>`_ - multi-tenant logging system
-* `PierreRochard/postgrest-boilerplate <https://github.com/PierreRochard/postgrest-boilerplate>`_ - example auth back-end
-* `marmelab/ng-admin-postgrest <https://github.com/marmelab/ng-admin-postgrest>`_ - automatic database admin panel
-* `seveibar/postgrest-vercel <https://github.com/seveibar/postgrest-vercel>`_ - run PostgREST on Vercel (Serverless/AWS Lambda)
+* `PostGUI <https://github.com/priyank-purohit/PostGUI>`_ - React Material UI admin panel
+* `PostgrestSkeleton <https://github.com/Recmo/PostgrestSkeleton>`_ - Docker Compose, PostgREST, Nginx and Auth0
+* `blogdemo <https://github.com/ruslantalpa/blogdemo>`_ - blog api demo in a vagrant image
```
steve-chavez, line 24: :O, where these automatically reordered?

wolfgangwalther: No, that was me hitting `:sort` in vim ;)

I gave them better names first (so removed the owner of the repo) to actually order by name.

wolfgangwalther: Butt.... I see what you mean. I should probably order case-insensitively... :D

install.rst
```diff
@@ -30,17 +33,82 @@ If you use **Nix**, then you can install PostgREST from nixpkgs.
...
+   - `--install-ghc` flag is only needed for the first build and can be omitted in the subsequent builds.
 
-Running
--------
+* Check that the server is installed: :code:`postgrest --help`.
+
+To properly run the test suite, you need a PostgreSQL database that the tests can run against. There are several ways to set up this database.
+
+Testing with a temporary database
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
steve-chavez, line 88: I think this part should be removed as well. It seems off from the "Installation" page, which should be sort of a quick read.

Additionally, I haven't been using stack for a long time now, only nix :-]

install.rst
```diff
@@ -30,17 +33,82 @@ If you use **Nix**, then you can install PostgREST from nixpkgs.
...
+
+If you have PostgreSQL installed locally (:code:`initdb`, :code:`pg_ctl` and :code:`psql` should be on your PATH, no server needs to be running), you can run the test suite against a temporary database:
+
+.. code:: bash
+
+   test/with_tmp_db stack test
+
+The :code:`with_tmp_db` script will set up a new PostgreSQL cluster in a temporary directory, set the required environment variables and run the command that you passed it as an argument, :code:`stack test` in the example above. When the command is done, the temporary database is torn down and deleted again.
+
+Testing with Docker
```
steve-chavez, line 98: Ditto. Remove as well.

install.rst
```diff
@@ -30,17 +33,82 @@ If you use **Nix**, then you can install PostgREST from nixpkgs.
...
 
 .. code:: bash
 
   choco install postgrest
   scoop install postgrest
 
-When a pre-built binary does not exist for your system you can :ref:`build the project from source <build_source>`.
+.. _build_source:
+
+Building from Source
```
steve-chavez, line 45: How about moving this to the bottom? Before the "Deploying-to-heroku" section. That way the Installation page is more straightforward.

index.rst
```diff
@@ -273,7 +260,7 @@ Testimonials
 
   -- Anupam Garg, Datrium, Inc.
 
-Translations
+Contributing
 ------------
 
-* `Chinese <http://postgrest.org/zh/latest/>`_ (latest version ``v0.4.2.0``)
+Please see the `Contributing guidelines <https://github.com/PostgREST/postgrest/blob/main/.github/CONTRIBUTING.md>`_ in the main PostgREST repository.
```
steve-chavez, line 266: Cool. I think we should leave all the dev docs on the main repo.

Btw, I've just noticed that on CONTRIBUTING.md we have links to:

- https://postgrest.org/en/stable/development.html#running-the-test-suite
- https://postgrest.org/en/stable/development.html#build-from-source

I think those can be removed after this PR is done.

steve-chavez:
2 and 4 LGTM! I've left some comments above, feel free to merge after you're done with those.

:+1: for linkcheck

PR #380, opened 2020-12-08
Fix outdated environment variable for docker compose

### LINKED ISSUE #379: PGRST_SERVER_PROXY_URI in docker-compose instructions is incorrect
I'm very new to PostgREST, so this might be wrong, but I think the instructions for the docker-compose install at https://postgrest.org/en/v7.0.0/install.html#docker are incorrect. In that compose file example, it specifies `PGRST_SERVER_PROXY_URI` should be http://127.0.0.1:3000. Regardless of what I put in this field, my openAPI response always showed the host as "0.0.0.0". After some poking around the code, it looks like this environment variable should actually be `PGRST_OPENAPI_SERVER_PROXY_URI` instead.

If someone can confirm this, I'm happy to make a PR to fix it in the docs, but I want to make sure this is the correct thing to do, first.

jat255:
Also removed a deprecated `links` setting from the docker-compose, since that's not needed any more

Resolves #379

### DIFF
--- a/install.rst
+++ b/install.rst
@@ -172,13 +172,11 @@ To avoid having to install the database at all, you can run both it and the serv
       image: postgrest/postgrest
       ports:
         - "3000:3000"
-      links:
-        - db:db
       environment:
         PGRST_DB_URI: postgres://app_user:password@db:5432/app_db
         PGRST_DB_SCHEMA: public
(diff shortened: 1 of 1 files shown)


wolfgangwalther:
> Also removed a deprecated `links` setting from the docker-compose, since that's not needed any more

Absolutely right, thanks!

PR #385, opened 2021-01-13
Change port in configuration of db-uri to default PostgreSQL port (5432).

Oleg2210:

### DIFF
--- a/tutorials/tut0.rst
+++ b/tutorials/tut0.rst
@@ -161,7 +161,7 @@ PostgREST uses a configuration file to tell it how to connect to the database. C
 
 .. code-block:: ini
 
-  db-uri = "postgres://authenticator:mysecretpassword@localhost:5433/postgres"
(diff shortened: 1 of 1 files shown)


tutorials/tut0.rst
```diff
@@ -161,7 +161,7 @@ PostgREST uses a configuration file to tell it how to connect to the database. C
 
 .. code-block:: ini
 
-  db-uri = "postgres://authenticator:mysecretpassword@localhost:5433/postgres"
+  db-uri = "postgres://authenticator:mysecretpassword@localhost:5432/postgres"
```
steve-chavez, line 164: Hi Oleg,

On a previous step we install pg in docker, which has the 5433 port assigned:
https://postgrest.org/en/v7.0.0/tutorials/tut0.html#step-2-install-postgresql

I'm guessing you didn't use pg for docker and instead used your system postgres?

5433 was used because users kept reporting conflicts with their system postgres on 5432.

PR #390, opened 2021-02-22
Fix broken link for 5min tutorial

timlinux:


steve-chavez:
Thanks! Do note that there are newer versions of the docs:

- https://postgrest.org/en/v7.0.0/ecosystem.html
- https://postgrest.org/en/latest/ecosystem.html

PR #391, opened 2021-03-12
Update api.rst

apanagio:
Changed SET LOCAL to SELECT set_config( ... ) so that the value of the header can be dynamically calculated. Supported in versions 9.6 and above.


steve-chavez:
Cool! Thank you @apanagio.

PR #395, opened 2021-04-06
change Master to Main in heroku deploy steps

wolfgangwalther:
I have never used Heroku. Is this "Master" referencing the default github branch? Then it should be changed as suggested.


steve-chavez (approved):
Yes, I think the change is good. Been a while since I've used Heroku.

Issue #335, opened 2020-06-24
Add docs for NOTIFY reload


steve-chavez:
References for https://github.com/PostgREST/postgrest/pull/1542. In the schema reloading and config pages.

We also need a recommended pg event trigger, see https://github.com/PostgREST/postgrest/issues/1512#issuecomment-643687504.

Issue #147, opened 2018-05-30
Rename Stored Procedures to Stored Functions


steve-chavez:
PostgreSQL 11 supports `CREATE PROCEDURE` see https://www.postgresql.org/docs/11/static/xproc.html, change the terms to be more precise in the next version.

steve-chavez:
Actually the proper term for generalizing procedure/function is "routine" I believe. But most people will look for "stored procedures".

Pg 11 stored procedures are not supported since they don't play well with transactions for now, see https://github.com/PostgREST/postgrest/issues/1210.

Perhaps we could add a note about this.

Videl:
Would you like a PR for this? It was pretty confusing for it to return: To call a procedure, use CALL

The documentation made me think it was supported because there is a part about stored procedures: http://postgrest.org/en/v7.0.0/api.html#stored-procedures

Would this documentation change be good: renaming the section of the link above + adding a warning that procedures are not supported because of details about transactions

Also, I wanted to use procedures because I though procedures were wrapped in a transaction.
I read in issue #1210 that there is a transaction through postgrest: would that revert inserts made in the function if the function fails?

steve-chavez:
@Videl I think a warning about pg 11 stored procedures would be nice. However I'm not sure about renaming the section because many people conflate pg procedures with pg functions.

> I read in issue #1210 that there is a transaction through postgrest: would that revert inserts made in the function if the function fails?

Yes, the insertions would be ROLLBACKed.

PR #404, opened 2021-06-01
Reorganize the Schema Cache information into a separate reference page

### LINKED ISSUE #352: Make schema cache a reference page
The schema cache is core to postgrest, but docs on this are a bit hidden on the admin page: http://postgrest.org/en/v7.0.0/admin.html#schema-reloading.

So the schema cache should have its own reference page.

laurenceisla:
Closes #352

WIP. Maybe more information needs to be added now that it's in a separate reference page?


admin.rst
```diff
@@ -196,58 +196,10 @@ Restart the database and watch the log file in real-time to understand how HTTP
...
-     EXECUTE PROCEDURE public.notify_ddl_postgrest();
-
-Then run the `pg_listen <https://github.com/begriffs/pg_listen>`_ utility to monitor for that event and send a SIGUSR1 when it occurs:
-
-.. code-block:: bash
-
-  pg_listen <db-uri> ddl_command_end $(which killall) -SIGUSR1 postgrest
-
-Now, whenever the structure of the database schema changes, PostgreSQL will notify the ``ddl_command_end`` channel, which will cause ``pg_listen`` to send PostgREST the signal to reload its cache. Note that pg_listen requires full path to the executable in the example above.
+If the schema changes while the server is running it results in a stale cache and leads to errors. To learn how to refresh the cache see :ref:`schema_reloading`.
```
steve-chavez, line 202:
```suggestion
Changing the schema while the server is running can lead to errors due to a stale schema cache. To learn how to refresh the cache see :ref:`schema_reloading`.
```

schema_cache.rst
```diff
@@ -0,0 +1,57 @@
...
+
+Schema Cache
+============
+
+Users are often confused by PostgREST's database schema cache. It is present because detecting foreign key relationships between tables (including how those relationships pass through views) is necessary, but costly. API requests consult the schema cache as part of :ref:`resource_embedding`. However if the schema changes while the server is running it results in a stale cache and leads to errors claiming that no relations are detected between tables.
+
+.. important::
+
+   Since v5.0, PostgREST also makes use of the schema cache for stored functions metadata: parameters, return type, volatility.
+   It also uses the schema cache for resolving overloaded functions. You should refresh the cache if a change in any of the prior is done.
```
steve-chavez, line 11: I think this note should be merged with the above paragraph now. It's been a while since v5.0.

steve-chavez:
> Maybe more information needs to be added now that it's in a separate reference page?

@laurenceisla Yeah, for one it should be really clear about why we need the schema cache.

Wolfgang identified the needs on https://github.com/PostgREST/postgrest/pull/1702

> I identified 3 core features where postgrest core relies on the schema cache: embedding relations, calling RPCs and creating Location headers after INSERT

I also mentioned some other things the schema cache needs there.

schema_cache.rst
```diff
@@ -0,0 +1,57 @@
...
+  CREATE EVENT TRIGGER ddl_postgrest ON ddl_command_end
+     EXECUTE PROCEDURE public.notify_ddl_postgrest();
+
+Then run the `pg_listen <https://github.com/begriffs/pg_listen>`_ utility to monitor for that event and send a SIGUSR1 when it occurs:
+
+.. code-block:: bash
+
+  pg_listen <db-uri> ddl_command_end $(which killall) -SIGUSR1 postgrest
+
+Now, whenever the structure of the database schema changes, PostgreSQL will notify the ``ddl_command_end`` channel, which will cause ``pg_listen`` to send PostgREST the signal to reload its cache. Note that pg_listen requires full path to the executable in the example above.
```
wolfgangwalther, line 57: I don't think we need `pg_listen` anymore, but can `NOTIFY` postgrest directly now?

laurenceisla: Yes. There's another issue https://github.com/PostgREST/postgrest-docs/issues/335 for this part of the documentation, though. I'll make another PR for it.

schema_cache.rst
```diff
@@ -0,0 +1,78 @@
...
+- For :ref:`s_procs`, it needs access to their metadata, for example, parameters, return type or volatility. It also needs to verify if the function is overloaded.
+- To do an :ref:`upsert`, it looks for the primary keys.
+- To do an INSERT, it also looks for the primary keys in order to return the Location header.
+- For OPTION requests, it verifies the existence and the data modification capabilities of a table or view.
+- For :ref:`open-api`, it needs information about objects in the database (e.g., object description, columns and parameters).
+
+Retrieving this information directly from the database is costly, that is why PostgREST uses a database schema cache instead. When you change any of the information mentioned above while PostgREST is running, the schema cache turns stale. You then need to reload the schema before you make a request related to these changes, otherwise you'll receive an error instead of the expected response.
+
+For instance, suppose you add a ``cities`` table to your database. This table has a foreign key ``country_id`` referencing an existing ``countries`` table. Then, you make a request to get the data from ``cities`` and their ``countries``:
```
steve-chavez, line 17: For the stale schema cache case, we can add another subheading. That way we add some emphasis and we can link to the section.

I suggest to split the paragraph like this:

```suggestion
Retrieving this information directly from the database is costly, that is why PostgREST uses a database schema cache instead.

The stale schema cache
--------------------------------

When you change any of the information mentioned above while PostgREST is running, the schema cache turns stale. You then need to reload the schema before you make a request related to these changes, otherwise you'll receive an error instead of the expected response.
```

steve-chavez: Also check the important note here: https://postgrest.org/en/latest/api.html#stored-procedures

The example with the function error should be here as well. That will no longer be true when https://github.com/PostgREST/postgrest/pull/1841 is merged, but for now it can be moved.

schema_cache.rst
```diff
@@ -0,0 +1,78 @@
+.. _schema_cache:
+
+Schema Cache
+============
+
+PostgREST uses the database schema cache to get information for several of its features:
+
+- For :ref:`resource_embedding`, it needs relationships' information, such as related tables, foreign keys, cardinalities, etc.
```
steve-chavez, line 8: I think we should be precise here, so no "etc". Also "cardinalities" is more of an internal abstraction that's used in the code, so it shouldn't be included either.

```suggestion
- For :ref:`resource_embedding`, it looks for foreign keys constraints to determine relationships between tables or views.
```

schema_cache.rst
```diff
@@ -0,0 +1,78 @@
+.. _schema_cache:
+
+Schema Cache
+============
+
+PostgREST uses the database schema cache to get information for several of its features:
+
+- For :ref:`resource_embedding`, it needs relationships' information, such as related tables, foreign keys, cardinalities, etc.
+- For :ref:`s_procs`, it needs access to their metadata, for example, parameters, return type or volatility. It also needs to verify if the function is overloaded.
```
steve-chavez, line 9: "For example" also sounds like lack of precision. So maybe like:

```suggestion
- For :ref:`s_procs`, it needs access to their metadata: parameters, return type, volatility, `overloading https://www.postgresql.org/docs/current/xfunc-overload.html`_.

```

schema_cache.rst
```diff
@@ -0,0 +1,78 @@
...
+============
+
+PostgREST uses the database schema cache to get information for several of its features:
+
+- For :ref:`resource_embedding`, it needs relationships' information, such as related tables, foreign keys, cardinalities, etc.
+- For :ref:`s_procs`, it needs access to their metadata, for example, parameters, return type or volatility. It also needs to verify if the function is overloaded.
+- To do an :ref:`upsert`, it looks for the primary keys.
+- To do an INSERT, it also looks for the primary keys in order to return the Location header.
+- For OPTION requests, it verifies the existence and the data modification capabilities of a table or view.
+- For :ref:`open-api`, it needs information about objects in the database (e.g., object description, columns and parameters).
```
steve-chavez, line 13: Ditto for "e.g.":

```suggestion
- For :ref:`open-api`, it looks for `comments <https://www.postgresql.org/docs/current/sql-comment.html>`_ on the database objects.
```

"Database objects" could be further clarified. I don't recall precisely which objects comments we look for.

schema_cache.rst
```diff
@@ -0,0 +1,78 @@
...
+Schema Cache
+============
+
+PostgREST uses the database schema cache to get information for several of its features:
+
+- For :ref:`resource_embedding`, it needs relationships' information, such as related tables, foreign keys, cardinalities, etc.
+- For :ref:`s_procs`, it needs access to their metadata, for example, parameters, return type or volatility. It also needs to verify if the function is overloaded.
+- To do an :ref:`upsert`, it looks for the primary keys.
+- To do an INSERT, it also looks for the primary keys in order to return the Location header.
+- For OPTION requests, it verifies the existence and the data modification capabilities of a table or view.
```
steve-chavez, line 12: Ditto for "data modification capabitilies". "instead of triggers" of the view would be more precise.

schema_cache.rst
laurenceisla: Looking at [the code](https://github.com/PostgREST/postgrest/blob/1b12b112a1e491f51e48e37d5d828e4a390de26b/src/PostgREST/App.hs#L462-L469), the comments are accessed directly from the database. Tested it and there's no need to reload the schema cache to get the comments, but columns, PK, FK, INSTEAD OF triggers, and procedures metadata do need it.

schema_cache.rst
laurenceisla: Agree, fixed the vague examples for this and the other comments.

steve-chavez:
@laurenceisla Once you've finished editing the schema cache page, you should add the change to [upcoming](https://postgrest--404.org.readthedocs.build/en/404/releases/upcoming.html). I've added docs improvements on releases before([on v7.0.0](https://postgrest--404.org.readthedocs.build/en/404/releases/v7.0.0.html), see "Documentation improvements"). That way users know we moved the "cache reloading" docs to another page.

schema_cache.rst
```diff
@@ -0,0 +1,111 @@
...
+.. code-block:: json
+
+ {
+  "hint":"No function matches the given name and argument types. You might need to add explicit type casts.",
+  "details":null,
+  "code":"42883",
+  "message":"function test.plus_one(num => text) does not exist"
+ }
+
+To solve these kind of issues you only need to reload the schema and repeat the request.
```
steve-chavez, line 65: I think we should remove "only". I've seen some articles(like this [one](https://www.knowledgeowl.com/home/dont-say-simply-jim-fisher)) that advice against mentioning that a task is "easy".

(We also had a similar case before, but I can't find it)

```suggestion
To solve these kind of issues you need to reload the schema and repeat the request.
```

steve-chavez, line 65: I think we should remove "only". I've seen some articles(like this [one](https://www.knowledgeowl.com/home/dont-say-simply-jim-fisher)) that advice against mentioning that a task is "easy".

(We had a complaint about this before, can't find the issue unfortunately)

```suggestion
To solve these kind of issues you need to reload the schema and repeat the request.
```

laurenceisla: Nice read! I agree.

schema_cache.rst
```diff
@@ -0,0 +1,111 @@
...
+- For INSERT statements, it also looks for the tables' primary keys in order to return the Location header.
+- For OPTION requests, it verifies the existence of tables and views and looks for INSTEAD OF triggers for the latter.
+- For :ref:`open-api`, it looks for information about tables (columns, primary keys and foreign keys), views (columns and INSTEAD OF triggers) and stored procedures (metadata).
+
+The Stale Schema Cache
+----------------------
+
+When you make a change in the database related to any of the features mentioned above while PostgREST is running, the schema cache turns stale. You then need to reload the schema before you make a request related to these changes, otherwise you'll receive an error instead of the expected response.
+
+For instance, suppose you add a ``cities`` table to your database. This table has a foreign key ``country_id`` referencing an existing ``countries`` table. Then, you make a request to get the data from ``cities`` and their ``countries``:
```
steve-chavez, line 22:
```suggestion

Stale foreign key relationships
~~~~~~~~~~~~~~~~~~~~~~~

For instance, suppose you add a ``cities`` table to your database. This table has a foreign key ``country_id`` referencing an existing ``countries`` table. Then, you make a request to get the data from ``cities`` and their ``countries``:
```

schema_cache.rst
```diff
@@ -0,0 +1,111 @@
...
+.. code-block:: json
+
+  {
+    "hint": "If a new foreign key between these entities was created in the database, try reloading the schema cache.",
+    "message": "Could not find a relationship between cities and countries in the schema cache"
+  }
+
+As you can see, PostgREST couldn't find the relationship in the schema cache.
+
+You get a similar result for stored procedures. Suppose you create this function on the database while PostgREST is running:
```
steve-chavez, line 39:
```suggestion

Stale function metadata
~~~~~~~~~~~~~~~~~~

You get a similar result for stored procedures. Suppose you create this function on the database while PostgREST is running:
```

steve-chavez: Also, let's change "procedures" for "functions" everywhere in this page. We actually don't support pg 11 stored procedures([ref](https://www.postgresql.org/docs/11/sql-createprocedure.html)), so this could cause confusion.

schema_cache.rst
```diff
@@ -0,0 +1,111 @@
...
+Schema Cache
+============
+
+There are several PostgREST features that need information from the database schema.
+If they accessed this information directly from the database every time they needed it, it would be too costly.
+That is why, when PostgREST starts, it generates a database schema cache and uses it to get the information needed for these features:
+
+- For :ref:`resource_embedding`, it looks for foreign key constraints to determine relationships between tables.
+- For :ref:`s_procs`, it looks for their metadata: parameters, return type, volatility and `overloading <https://www.postgresql.org/docs/current/xfunc-overload.html>`_.
+- For :ref:`upsert` statements, it looks for the tables' primary keys.
```
steve-chavez, line 12: Hm, let's not refer to "statements" here (in pg there's no UPSERT statement)

```suggestion
- For :ref:`upsert`, it looks for the tables' primary keys.
```

schema_cache.rst
```diff
@@ -0,0 +1,111 @@
...
+- For INSERT statements, it also looks for the tables' primary keys in order to return the Location header.
```
steve-chavez, line 13: This one should link to https://postgrest--404.org.readthedocs.build/en/404/api.html#insertions-updates

schema_cache.rst
```diff
@@ -0,0 +1,111 @@
...
+- For OPTION requests, it verifies the existence of tables and views and looks for INSTEAD OF triggers for the latter.
```
steve-chavez, line 14: Just FYI, I've opened https://github.com/PostgREST/postgrest-docs/issues/406 since we have no OPTIONS section.

schema_cache.rst
wolfgangwalther: Yes, fully agree with that one!

I think we also need to improve https://postgrest.org/en/latest/api.html#stored-procedures.

We should make it explicitly clear over there, that we don't support procedures, but functions - even when we name the concept a "remote procedure call - RPC". Not sure if we should use the term "stored procedures" at all, though.

laurenceisla: > Also, let's change "procedures" for "functions" everywhere in this page. We actually don't support pg 11 stored procedures(ref), so this could cause confusion.

Oh, I get it. But does this mean that the reference about [stored procedures](https://postgrest.org/en/v7.0.0/api.html#stored-procedures) should be renamed in the future?

schema_cache.rst
```diff
@@ -0,0 +1,119 @@
...
+For instance, let's see what would happen if you have a stale schema for foreign key relationships and function metadata:
+
+Stale Foreign Key Relationships
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+    Suppose you add a ``cities`` table to your database. This table has a foreign key ``country_id`` referencing an existing ``countries`` table. Then, you make a request to get the data from ``cities`` and their ``countries``:
+
+    .. code-block:: http
+
+      GET /cities?select=name,country:countries(id,name) HTTP/1.1
```
steve-chavez, line 31: @laurenceisla This section is showing as indented. I don't think it should, it looks a bit weird. See https://postgrest--404.org.readthedocs.build/en/404/schema_cache.html#stale-foreign-key-relationships.

schema_cache.rst
```diff
@@ -0,0 +1,119 @@
...
+        "hint": "If a new foreign key between these entities was created in the database, try reloading the schema cache.",
+        "message": "Could not find a relationship between cities and countries in the schema cache"
+      }
+
+    As you can see, PostgREST couldn't find the relationship in the schema cache.
+
+Stale Function Metadata
+~~~~~~~~~~~~~~~~~~~~~~~
+
+    Suppose you create this function on the database while PostgREST is running:
```
steve-chavez, line 47: Ditto for this one - indented.

schema_cache.rst
laurenceisla: True that, the indentation doesn't look good. It's fixed now.

schema_cache.rst
```diff
@@ -0,0 +1,119 @@
...
+There are several PostgREST features that need information from the database schema.
+If they accessed this information directly from the database every time they needed it, it would be too costly.
+That is why, when PostgREST starts, it generates a database schema cache and uses it to get the information needed for these features:
+
+- For :ref:`resource_embedding`, it looks for foreign key constraints to determine relationships between tables.
+- For :ref:`Database Functions <s_procs>`, it looks for their metadata: parameters, return type, volatility and `overloading <https://www.postgresql.org/docs/current/xfunc-overload.html>`_.
+- For :ref:`Upserts <upsert>`, it looks for the tables' primary keys.
+- For :ref:`insert_update`, it also looks for the tables' primary keys in order to return the Location header.
+- For OPTIONS requests, it verifies the existence of tables and views and looks for INSTEAD OF triggers for the latter.
+- For :ref:`open-api`, it looks for information about tables (columns, primary keys and foreign keys), views (columns and INSTEAD OF triggers) and functions (metadata).
```
steve-chavez, line 15: @laurenceisla Do you think this part would look better as a table? We might be able to reduce some repeated words by putting a column for "feature" and another column for "required metadata".

laurenceisla: I agree, it should make it look less cluttered.

laurenceisla: > ... and another column for "required metadata".

I used "required schema cache information" instead, I think that way it should include everything in the list, WDYT?

schema_cache.rst
```diff
@@ -0,0 +1,135 @@
...
+
+Schema Cache
+============
+
+There are several PostgREST features that need information from the database schema.
+If they accessed this information directly from the database every time they needed it, it would be too costly.
+That is why, when PostgREST starts, it generates a database schema cache and uses it to get the information needed for these features:
+
++-------------------------------------+-------------------------------------------------------------------------------+
+| PostgREST Feature                   | Required Schema Cache Information                                             |
```
steve-chavez, line 11: I think only "feature" should do:

```suggestion
|  Feature                   | Required Schema Cache Information                                             |
```

schema_cache.rst
steve-chavez: @laurenceisla The table looks much better!

Regarding "required schema cache information", I think "required metadata" is more precise since the metadata is not always in the cache, the schema cache might have to be reloaded to get it.

schema_cache.rst
steve-chavez: > Oh, I get it. But does this mean that the reference about stored procedures should be renamed in the future?

Yeah, there's an issue related to that: https://github.com/PostgREST/postgrest-docs/issues/147. For now, we can use "function" instead of "procedure" on the schema cache page.

schema_cache.rst
laurenceisla: > Regarding "required schema cache information", I think "required metadata" is more precise since the metadata is not always in the cache, the schema cache might have to be reloaded to get it.

Ah, I get it now. I agree.

schema_cache.rst
```diff
@@ -0,0 +1,131 @@
...
+.. code-block:: json
+
+  {
+    "hint": "If a new foreign key between these entities was created in the database, try reloading the schema cache.",
+    "message": "Could not find a relationship between cities and countries in the schema cache"
+  }
+
+As you can see, PostgREST couldn't find the relationship in the schema cache. See the section :ref:`schema_reloading` to solve this issue.
+
+Stale Function Metadata
```
steve-chavez, line 56: [Signature](https://developer.mozilla.org/en-US/docs/Glossary/Signature/Function) might be a more precise term.
```suggestion
Stale Function Signature
```

schema_cache.rst
```diff
@@ -0,0 +1,131 @@
...
+----------------------
+
+When you make changes on the metadata mentioned above while PostgREST is running, the schema cache turns stale. Future requests you make related to these changes will need the :ref:`schema to be reloaded <schema_reloading>`; otherwise, you'll get an error instead of the expected result.
+
+For instance, let's see what would happen if you have a stale schema for foreign key relationships and function metadata:
+
+Stale Foreign Key Relationships
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Suppose you add a ``cities`` table to your database. This table has a foreign key ``country_id`` referencing an existing ``countries`` table. Then, you make a request to get the data from ``cities`` and their ``countries``:
```
steve-chavez, line 39:
```suggestion
Suppose you add a ``cities`` table to your database. This table has a foreign key referencing an existing ``countries`` table. Then, you make a request to get the ``cities`` and their belonging ``countries``:
```

schema_cache.rst
```diff
@@ -0,0 +1,131 @@
...
+
+.. code-block:: http
+
+  GET /cities?select=name,country:countries(id,name) HTTP/1.1
+
+But instead, you get an error message that looks like this:
```
steve-chavez, line 45:
```suggestion
But instead, you'll get an error message that looks like this:
```

schema_cache.rst
```diff
@@ -0,0 +1,131 @@
...
+
+.. code-block:: json
+
+  {
+    "hint": "If a new foreign key between these entities was created in the database, try reloading the schema cache.",
+    "message": "Could not find a relationship between cities and countries in the schema cache"
+  }
+
+As you can see, PostgREST couldn't find the relationship in the schema cache. See the section :ref:`schema_reloading` to solve this issue.
```
steve-chavez, line 54:
```suggestion
As you can see, PostgREST couldn't find the newly created foreign key in the schema cache. See the section :ref:`schema_reloading` to solve this issue.
```

schema_cache.rst
```diff
@@ -0,0 +1,131 @@
...
+
+Stale Function Metadata
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Suppose you create this function on the database while PostgREST is running:
```
steve-chavez, line 59:
```suggestion
Suppose you create the following function while PostgREST is running:
```

schema_cache.rst
```diff
@@ -0,0 +1,131 @@
...
++                                            +-------------------------------------------------------------------------------+
+|                                            | View columns and INSTEAD OF TRIGGERS                                          |
++                                            +-------------------------------------------------------------------------------+
+|                                            | Function's metadata                                                           |
++--------------------------------------------+-------------------------------------------------------------------------------+
+
+The Stale Schema Cache
+----------------------
+
+When you make changes on the metadata mentioned above while PostgREST is running, the schema cache turns stale. Future requests you make related to these changes will need the :ref:`schema to be reloaded <schema_reloading>`; otherwise, you'll get an error instead of the expected result.
```
steve-chavez, line 32:
```suggestion
When you make changes on the metadata mentioned above while PostgREST is running, the schema cache turns stale. Future requests that use the above features will need the :ref:`schema to be reloaded <schema_reloading>`; otherwise, you'll get an error instead of the expected result.
```

schema_cache.rst
```diff
@@ -0,0 +1,131 @@
...
+============
+
+Certain PostgREST features need information from the database schema; however, accessing the database directly would be too costly, that is why PostgREST generates a database schema cache on startup and gets the information from it instead:
+
++--------------------------------------------+-------------------------------------------------------------------------------+
+| Feature                                    | Required Metadata                                                             |
++============================================+===============================================================================+
+| :ref:`resource_embedding`                  | Foreign key constraints (to determine relationships between tables)           |
++--------------------------------------------+-------------------------------------------------------------------------------+
+| :ref:`Database Functions <s_procs>`        | Function's metadata (parameters, return type, volatility and                  |
```
steve-chavez, line 13:
```suggestion
| :ref:`Database Functions <s_procs>`        | Function metadata: parameters, return type, volatility and                  |
```

schema_cache.rst
```diff
@@ -0,0 +1,131 @@
...
+|                                            | `overloading <https://www.postgresql.org/docs/current/xfunc-overload.html>`_) |
++--------------------------------------------+-------------------------------------------------------------------------------+
+| :ref:`Upserts <upsert>`                    | Primary keys                                                                  |
++--------------------------------------------+-------------------------------------------------------------------------------+
+| :ref:`insert_update`                       | Primary keys (in order to return the Location header)                         |
```
steve-chavez, line 18: Only insertions need the pk.
```suggestion
| :ref:`Insertions <insert_update>`                       | Primary keys (in order to return the Location header)                         |
```

steve-chavez: @laurence Btw, the Location header is optional now(`return=headers`). Perhaps this need to be cleared here.

steve-chavez: @laurenceisla Btw, the Location header is optional now(return=headers). Perhaps we should clear this here.

schema_cache.rst
```diff
@@ -0,0 +1,131 @@
...
++--------------------------------------------+-------------------------------------------------------------------------------+
+| :ref:`OPTIONS requests <options_requests>` | View primary key columns and INSTEAD OF TRIGGERS                              |
```
steve-chavez, line 20:
```suggestion
| :ref:`OPTIONS requests <options_requests>` | View's INSTEAD OF TRIGGERS and primary keys                         |
```

schema_cache.rst
```diff
@@ -0,0 +1,131 @@
...
++--------------------------------------------+-------------------------------------------------------------------------------+
+| :ref:`open-api`                            | Table columns, primary keys and foreign keys                                  |
++                                            +-------------------------------------------------------------------------------+
+|                                            | View columns and INSTEAD OF TRIGGERS                                          |
++                                            +-------------------------------------------------------------------------------+
+|                                            | Function's metadata                                                           |
```
steve-chavez, line 26: Ah, there used to be a mention about comments on these db objects right? I think we should add them again.

schema_cache.rst
laurenceisla: In this case, I think the sentence follows the present tense of the preceding one:

"Then, you make a request to get the ``cities`` and their belonging ``countries`` ... but instead, you get an error..."

schema_cache.rst
laurenceisla: Yes, but I took them away because comments don't depend on the schema cache. I explained it better [in this comment](https://github.com/PostgREST/postgrest-docs/pull/404#discussion_r647051537)

steve-chavez: Oops, I forgot about that. Thanks.

schema_cache.rst
```diff
@@ -0,0 +1,131 @@
...
++--------------------------------------------+-------------------------------------------------------------------------------+
+
+The Stale Schema Cache
+----------------------
+
+When you make changes on the metadata mentioned above while PostgREST is running, the schema cache turns stale. Future requests that use the above features will need the :ref:`schema to be reloaded <schema_reloading>`; otherwise, you'll get an error instead of the expected result.
```
steve-chavez, line 32:
```suggestion
When you make changes on the metadata mentioned above, the schema cache will turn stale on a running PostgREST. Future requests that use the above features will need the :ref:`schema to be reloaded <schema_reloading>`; otherwise, you'll get an error instead of the expected result.
```

schema_cache.rst
```diff
@@ -0,0 +1,131 @@
+.. _schema_cache:
+
+Schema Cache
+============
+
+Certain PostgREST features need information from the database schema; however, accessing the database directly would be too costly, that is why PostgREST generates a database schema cache on startup and gets the information from it instead:
```
steve-chavez, line 6:
```suggestion
PostgREST caches metadata from the database schema to avoid repeating expensive queries. This metadata is not required by all of the PostgREST features, only the following:
```

steve-chavez (changes requested):
I think this should be the last review for this one. It's mostly rewording.

schema_cache.rst
```diff
@@ -0,0 +1,131 @@
...
+  "details":null,
+  "code":"42883",
+  "message":"function test.plus_one(num => text) does not exist"
+ }
+
+See the section :ref:`schema_reloading` to solve this issue.
+
+.. _schema_reloading:
+
+Schema Reloading
```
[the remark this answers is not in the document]
laurenceisla: I think the link should stay that way because we've referenced it many times in answers to users' issues. I don't think it's that inaccurate as a reference name either.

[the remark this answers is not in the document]
steve-chavez: Yeah, you're right. Too late to change it. The name is not so bad, let's leave it like that.

steve-chavez (approved):

PR #407, opened 2021-06-11
Add an option section to API reference page

### LINKED ISSUE #406: Add section for OPTIONS
An example of the OPTIONS http method. Also how it considers VIEWs instead of triggers.

laurenceisla:
Closes #406

### DIFF
--- a/api.rst
+++ b/api.rst
@@ -1454,6 +1454,73 @@ You can use a tool like `Swagger UI <https://swagger.io/tools/swagger-ui/>`_ to 
 
 .. _multiple-schemas:
 
+OPTIONS requests
+================
+
+You can verify which HTTP methods are allowed on endpoints for tables and views using an OPTIONS request.
+
+Tables
+------
+
+Suppose you have the ``people`` table. When you make this API request:
+
+.. code-block:: http
+
+  OPTIONS /people HTTP/1.1
+
+The response you get is:
+
+.. code-block:: http
+
+  HTTP/1.1 200 OK
+  Allow: OPTIONS,GET,HEAD,POST,PUT,PATCH,DELETE
+
+Notice that for the ``people`` table endpoint (and for any table endpoint in general), all those methods *can* be used even though the user
+making the request doesn't have full permissions on the table. In other words, it verifies what *can* be done on the table, not what permissions
+the user has on it. In this case, for example, even if the database user only has a reading
+permission on the ``people`` table, the OPTIONS request will give the same response as the example above. The user can then do a
+a POST request on the table resulting in a ``401 Unauthorized`` status response.
+
+Views
+-----
+
+The user's permission concept for tables is the same for views, but there's a difference in verifying what operations *can* be done on them.
+If the view is `auto-updatable <https://www.postgresql.org/docs/current/sql-createview.html#SQL-CREATEVIEW-UPDATABLE-VIEWS>`_, then all methods are allowed;
+otherwise, the methods that can be used are determined by the view's INSTEAD OF TRIGGERS and primary key columns:
+
+=================== ==========================================================================
+ Method allowed      View's requirements
+=================== ==========================================================================
+ OPTIONS, GET, HEAD  None (Always allowed)
+ POST                INSTEAD OF INSERT TRIGGER
+ PUT                 INSTEAD OF INSERT TRIGGER, INSTEAD OF UPDATE TRIGGER, primary key columns
+ PATH                INSTEAD OF UPDATE TRIGGER
+ DELETE              INSTEAD OF DELETE TRIGGER
+=================== ==========================================================================
+
+For instance, suppose you have a **non** `auto-updatable` view named ``people_view`` that has an INSTEAD OF INSERT and an INSTEAD OF DELETE trigger.
+When you make this request:
+
+.. code-block:: http
+
+  OPTIONS /people_view HTTP/1.1
+
+The response you get is:
+
+.. code-block:: http
+
+  HTTP/1.1 200 OK
+  Allow: OPTIONS,GET,HEAD,POST,DELETE
+
+Functions
+~~~~~~~~~
+
+For database functions endpoints, OPTIONS requests are not supported.
+
+.. important::
+  Whenever you add or remove tables, views or views' INSTEAD OF TRIGGERS or modify tables' primary keys or views' definition on the database, you must refresh PostgREST's schema cache for OPTIONS requests to work properly. See the section :ref:`schema_reloading`.
+
+
 Switching Schemas
 =================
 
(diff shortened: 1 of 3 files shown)


steve-chavez:
I think we should include a CORS subsection here. There used to be one in a past version: https://postgrest.org/en/v3.2/api_reading.html#cors.

(@laurenceisla you could just copy that)

I'd say this should go first in the OPTIONS section, we've had some questions about this before.

api.rst
```diff
@@ -1454,6 +1454,73 @@ You can use a tool like `Swagger UI <https://swagger.io/tools/swagger-ui/>`_ to
 
 .. _multiple-schemas:
 
+OPTIONS requests
```
steve-chavez, line 1457: I think just "OPTIONS" would do.
```suggestion
OPTIONS
```

api.rst
```diff
@@ -1454,6 +1454,73 @@ You can use a tool like `Swagger UI <https://swagger.io/tools/swagger-ui/>`_ to
...
+
+.. code-block:: http
+
+  HTTP/1.1 200 OK
+  Allow: OPTIONS,GET,HEAD,POST,DELETE
+
+Functions
+~~~~~~~~~
+
+For database functions endpoints, OPTIONS requests are not supported.
```
steve-chavez, line 1518: Good to be explicit about this :+1:

api.rst
```diff
@@ -1452,6 +1452,68 @@ You can use a tool like `Swagger UI <https://swagger.io/tools/swagger-ui/>`_ to
...
+
+OPTIONS
+=======
+
+You can verify which HTTP methods are allowed on endpoints for tables and views using an OPTIONS request. These methods are allowed depending on what operations *can* be done on the table or view, not on the database permissions assigned to them.
+
+CORS
+----
+
+PostgREST sets highly permissive cross origin resource sharing, that is why it accepts Ajax requests from any domain.
```
steve-chavez, line 1465: Now that I gave it a second look. I think this should go last on the OPTIONS section - otherwise it breaks the flow of the text.

api.rst
```diff
@@ -1452,6 +1452,68 @@ You can use a tool like `Swagger UI <https://swagger.io/tools/swagger-ui/>`_ to
...
+
+Tables
+------
```
steve-chavez, line 1468: One issue with this heading, its URL fragment is simply "tables":
https://postgrest--407.org.readthedocs.build/en/407/api.html#tables

That might conflict with a future fragment. Maybe rename it to `options-tables`:
```suggestion

.. _options_tables:

Tables
------
```

api.rst
```diff
@@ -1452,6 +1452,68 @@ You can use a tool like `Swagger UI <https://swagger.io/tools/swagger-ui/>`_ to
...
+Then, the response you get is:
+
+.. code-block:: http
+
+  HTTP/1.1 200 OK
+  Allow: OPTIONS,GET,HEAD,POST,PUT,PATCH,DELETE
+
+
+Views
+-----
```
steve-chavez, line 1485: Ditto:
```suggestion

.. _options_views:

Views
-----
```

api.rst
```diff
@@ -1452,6 +1452,68 @@ You can use a tool like `Swagger UI <https://swagger.io/tools/swagger-ui/>`_ to
...
+
+The methods that can be used on views are determined by the presence of INSTEAD OF TRIGGERS and primary key columns:
```
steve-chavez, line 1487: I think we can leave the exception on the table row.

```suggestion
The methods that can be used on views are determined by the presence of INSTEAD OF TRIGGERS:
```

api.rst
```diff
@@ -1452,6 +1452,68 @@ You can use a tool like `Swagger UI <https://swagger.io/tools/swagger-ui/>`_ to
...
+   :widths: auto
+
+   +--------------------+-------------------------------------------------------------------------------------------------+
+   | Method allowed     | View's requirements                                                                             |
+   +====================+=================================================================================================+
+   | OPTIONS, GET, HEAD | None (Always allowed)                                                                           |
+   +--------------------+-------------------------------------------------------------------------------------------------+
+   | POST               | INSTEAD OF INSERT TRIGGER                                                                       |
+   +--------------------+-------------------------------------------------------------------------------------------------+
+   | PUT                | INSTEAD OF INSERT TRIGGER, INSTEAD OF UPDATE TRIGGER, primary key columns                       |
```
steve-chavez, line 1499:
```suggestion
   | PUT                | INSTEAD OF INSERT TRIGGER, INSTEAD OF UPDATE TRIGGER, also requires the presence of a  primary key                       |
```

api.rst
```diff
@@ -1452,6 +1452,68 @@ You can use a tool like `Swagger UI <https://swagger.io/tools/swagger-ui/>`_ to
...
+   | `auto-updatable views <https://www.postgresql.org/docs/current/sql-createview.html#SQL-CREATEVIEW-UPDATABLE-VIEWS>`_ |
+   +--------------------+-------------------------------------------------------------------------------------------------+
+
+Functions
+~~~~~~~~~
+
+For database function endpoints, OPTIONS requests are not supported.
+
+.. important::
+  Whenever you add or remove tables or views, or modify a view's definition, primary key columns or INSTEAD OF TRIGGERS on the database, you must refresh PostgREST's schema cache for OPTIONS requests to work properly. See the section :ref:`schema_reloading`.
```
steve-chavez, line 1515: Summarizing this a bit, since it's mostly about views
```suggestion
  Whenever you modify a view's INSTEAD OF TRIGGERS on the database, you must refresh PostgREST's schema cache for OPTIONS requests to work properly. See the section :ref:`schema_reloading`.
```

steve-chavez (commented):
Last review! Then it should be ready to merge :rocket:

api.rst
laurenceisla: > Summarizing this a bit, since it's mostly about views

I think at least a mention to table creation should be kept there. If a new table is created the OPTIONS result will be a 404 till the schema is reloaded, for instance.

steve-chavez: Cool, I agree. A mention to tables' existence should be kept.

api.rst
```diff
@@ -1452,6 +1452,75 @@ You can use a tool like `Swagger UI <https://swagger.io/tools/swagger-ui/>`_ to
 
   The OpenAPI information can go out of date as the schema changes under a running server. To learn how to refresh the cache see :ref:`schema_reloading`.
 
+.. _options_requests:
+
+OPTIONS
+=======
+
+You can verify which HTTP methods are allowed on endpoints for tables and views using an OPTIONS request. These methods are allowed depending on what operations *can* be done on the table or view, not on the database permissions assigned to them.
```
steve-chavez, line 1460:
```suggestion
You can verify which HTTP methods are allowed on endpoints for tables and views by using an OPTIONS request. These methods are allowed depending on what operations *can* be done on the table or view, not on the database permissions assigned to them.
```

Issue #114, opened 2017-11-05
Document how to do schema reloading on windows


begriffs:
Windows doesn't have a SIGHUP signal, and that's currently how you tell PostgREST to reload the schema. In the future when hasql supports LISTEN/NOTIFY we can listen for a db notification to trigger this. Until then @gcloeval contributed a batchfile to do it:

```batch
REM to run pass postgrest.exe CONFIG_FILE_NAME, ex: c:\workspace>restartpgrest.bat postgrest.exe pgrest.config
@if %1.==. (goto error) ELSE (goto restart)

:restart
@taskkill /f /im %1 >nul
@timeout /t 3 /nobreak >nul
@echo now starting %1
%1 %2
@echo restart complete
@goto exit

:error
@echo Please specify a program...
@goto exit

:exit
```

Add this to the docs about schema reloading.

laurenceisla:
@steve-chavez I think this was solved in https://github.com/PostgREST/postgrest-docs/pull/408#discussion_r676029939?

steve-chavez:
True. Closing.

PR #408, opened 2021-06-22
Add database notification functionality and configuration variables for schema reloading

### LINKED ISSUE #335: Add docs for NOTIFY reload

laurenceisla:
Closes #335


steve-chavez:
> Replace pg_listen

@laurenceisla I think we should leave the `pg_listen` option, but at the bottom. The thing about `db-channel` is that it needs a postgres session, and when using postgrest behind a connection pooler(like pgbouncer) that is not always possible.

configuration.rst
```diff
@@ -131,6 +133,20 @@ db-extra-search-path
 
   Multiple schemas can be added in a comma-separated string, e.g. ``public, extensions``.
 
+.. _db-channel:
+
+db-channel
+----------
+
+  The name of the notification channel that PostgREST uses to reload the schema cache. For more information, see :ref:`schema_reloading`.
```
steve-chavez, line 141: Note that the db-channel is also used for reloading the config. Check https://github.com/PostgREST/postgrest-docs/issues/409.

laurenceisla:
> > Replace pg_listen
> 
> @laurenceisla I think we should leave the `pg_listen` option, but at the bottom. The thing about `db-channel` is that it needs a postgres session, and when using postgrest behind a connection pooler(like pgbouncer) that is not always possible.

I've checked the [`pg_listen` utility](https://github.com/begriffs/pg_listen/blob/master/pg_listen.c#L190) and it also uses the `LISTEN` command on a notification channel, so it wouldn't work with a connection pooler either. In this case I think `pg_listen` is entirely replaced by the `NOTIFY` command.

steve-chavez:
@laurenceisla Good catch!

You're right, `pg_listen` would have the same problem, mentioning it as an alternative would make no sense.

We should notify our users of this change though. How about adding the change in the Upcoming page?

schema_cache.rst
```diff
@@ -106,26 +106,50 @@ To refresh the cache without restarting the PostgREST server, send the server pr
...
 
-The above is the manual way to do it. To automate cache reloads, use a database trigger like this:
+Another option is to send a `database notification event <https://www.postgresql.org/docs/current/sql-notify.html>`_ using the ``NOTIFY`` command. To enable this feature, you need to add these variables to the configuration file:
+
+.. code::
+
+  # postgrest.conf
+
+  # The name of the notification channel PostgREST will listen to
+  db-channel = pgrst
```
steve-chavez, line 116: `db-channel=pgrst` is set by default, I think we can omit this setting. `db-channel-enabled=true` is the only requisite.
```suggestion
```

configuration.rst
```diff
@@ -131,6 +133,20 @@ db-extra-search-path
...
+----------
+
+  The name of the notification channel that PostgREST uses to re read the config file and reload the schema cache. For more information, see :ref:`schema_reloading`.
+
+.. _db-channel-enabled:
+
+db-channel-enabled
+------------------
+
+  When this is set to :code:`true`, the notification channel specified in :ref:`db-channel` is enabled.
```
steve-chavez, line 148: We should mention here that this setting is not compatible with connection poolers using the transaction mode.

configuration.rst
```diff
@@ -131,6 +133,21 @@ db-extra-search-path
 
   Multiple schemas can be added in a comma-separated string, e.g. ``public, extensions``.
 
+.. _db-channel:
+
+db-channel
+----------
+
+  The name of the notification channel that PostgREST uses to re read the config file and reload the schema cache. For more information, see :ref:`schema_reloading`.
```
steve-chavez, line 141:
```suggestion
  The name of the notification channel that PostgREST uses for :ref:`schema_reloading` and configuration reloading.
```

configuration.rst
```diff
@@ -39,6 +39,8 @@ db-anon-role             String             Y
 db-pool                  Int     10
 db-pool-timeout          Int     10
 db-extra-search-path     String  public
+db-channel               String  pgrst
+db-channel-enabled       Boolean False
```
steve-chavez, line 43: @laurenceisla Now that I remember, we agreed somewhere that `db-channel-enabled` must be True by default. We wanted to be conservative because of pgbouncer, but we already have `db-prepared-statements` as True, which wouldn't work with pgbouncer anyway.

You could make this change in postgrest itself - change `db-channel-enabled` to True by default.

schema_cache.rst
```diff
@@ -106,26 +106,52 @@ To refresh the cache without restarting the PostgREST server, send the server pr
...
 
-The above is the manual way to do it. To automate cache reloads, use a database trigger like this:
+Another option is to send a `database notification event <https://www.postgresql.org/docs/current/sql-notify.html>`_. To enable this feature, add this variable to the configuration file:
+
+.. code::
+
+  # postgrest.conf
+
+  db-channel-enabled = true
+
```
steve-chavez, line 116: Mentioning the config parameter in the paragraph gives us the opportunity to link:
```suggestion
Another option is to send a `database notification event <https://www.postgresql.org/docs/current/sql-notify.html>`_. To enable this feature, set db-channel-enabled(link here) to true.
```

schema_cache.rst
```diff
@@ -106,26 +106,52 @@ To refresh the cache without restarting the PostgREST server, send the server pr
...
+  NOTIFY pgrst
+
+  -- Works the same way as:
+
+  NOTIFY pgrst, 'reload schema'
+
+
+.. important::
+
+   The ``NOTIFY`` command is not compatible with connection poolers like PgBouncer in transaction pooling mode.
```
steve-chavez, line 130: Ah, it's not the NOTIFY that's not compatible, it's the LISTEN that we do internally.

(could need rewording)
```suggestion
   db-channel is not compatible with connection poolers like PgBouncer in transaction pooling mode. Set it to false if you're using PostgREST behind a connection pooler.
```

laurenceisla: I used the [SQL feature map](https://www.pgbouncer.org/features.html#sql-feature-map-for-pooling-modes) from PgBouncer as reference, it mentions both `LISTEN/NOTIFY` as incompatible. However, I like the way you worded it, it includes the whole feature there.

configuration.rst
laurenceisla: Found the comment where you agreed to it  https://github.com/PostgREST/postgrest/issues/1622#issuecomment-733066644 and Wolfgang gives a very good explanation on to why it should be True by default. I'll make a PR in core and come back to the docs when the changes are merged.

configuration.rst
```diff
@@ -131,6 +133,21 @@ db-extra-search-path
...
+
+  The name of the notification channel that PostgREST uses for :ref:`schema_reloading` and configuration reloading.
+
+.. _db-channel-enabled:
+
+db-channel-enabled
+------------------
+
+  When this is set to :code:`true`, the notification channel specified in :ref:`db-channel` is enabled.
+  Note that this setting is not compatible with connection poolers like PgBouncer in transaction pooling mode.
```
steve-chavez, line 149: I think we should remove this for now, since we're not really compatible with pgbouncer.
```suggestion
```

schema_cache.rst
```diff
@@ -106,26 +106,44 @@ To refresh the cache without restarting the PostgREST server, send the server pr
...
+  NOTIFY pgrst
+
+  -- Works the same way as:
+
+  NOTIFY pgrst, 'reload schema'
+
+.. important::
+
+  The ``db-channel-enable`` :ref:`configuration parameter <db-channel-enabled>` enables the notification channel by default.
+  This setting is incompatible with connection poolers like PgBouncer in transaction pooling mode. Set it to ``false`` if you are using PostgREST behind a connection pooler.
```
steve-chavez, line 122: Ditto.
```suggestion
```

steve-chavez (changes requested):

schema_cache.rst
```diff
@@ -106,26 +106,43 @@ To refresh the cache without restarting the PostgREST server, send the server pr
      # or in docker-compose
      docker-compose kill -s SIGUSR1 <service>
 
-The above is the manual way to do it. To automate cache reloads, use a database trigger like this:
+Another option is to send a `database notification event <https://www.postgresql.org/docs/current/sql-notify.html>`_ from any PostgreSQL client by executing the ``NOTIFY`` command as follows:
```
steve-chavez, line 109: @laurenceisla I think we need some subheadings for making this more clear. So the first part for reloading would be about "Unix Signals" and the second about "NOTIFY"(postgres native IPC).

releases/upcoming.rst
```diff
@@ -16,6 +16,9 @@ Added
 * Allow :ref:`s_procs_variadic`.
   |br| -- `@wolfgangwalther <https://github.com/wolfgangwalther>`_
 
+* Allow schema cache reloading using the :ref:`schema_reloading_notify` command from any PostgreSQL client.
```
steve-chavez, line 19: Shortening a bit:

```suggestion
* Allow schema cache reloading using the PostgreSQL :ref:`schema_reloading_notify`.
```

schema_cache.rst
```diff
@@ -106,26 +111,52 @@ To refresh the cache without restarting the PostgREST server, send the server pr
      # or in docker-compose
      docker-compose kill -s SIGUSR1 <service>
 
-The above is the manual way to do it. To automate cache reloads, use a database trigger like this:
+.. _schema_reloading_notify:
+
+NOTIFY
+~~~~~~
+
+Send a `database notification <https://www.postgresql.org/docs/current/sql-notify.html>`_ from any PostgreSQL client by executing the ``NOTIFY`` command as follows:
```
steve-chavez, line 119:
```suggestion
There are environments where you can't easily send Unix Signals(like managed containers in cloud services). For this reason, PostgREST also allows you to reload its state through PostgreSQL `NOTIFY <https://www.postgresql.org/docs/current/sql-notify.html>`_ as follows:

File: schema_cache.rst [hunk 13]
```diff
@@ -106,26 +111,52 @@ To refresh the cache without restarting the PostgREST server, send the server pr
...
 .. code-block:: postgresql
 
-  CREATE OR REPLACE FUNCTION public.notify_ddl_postgrest()
-    RETURNS event_trigger
-   LANGUAGE plpgsql
+  NOTIFY pgrst
+
+  -- Works the same way as:
+
+  NOTIFY pgrst, 'reload schema'
```
steve-chavez, line 127: I'd prefer to be explicit here, the empty body NOTIFY might change in the future.
```suggestion
  NOTIFY pgrst, 'reload schema'
```

steve-chavez, line 127: The empty body NOTIFY might change in the future, it'd be better to be explicit here.
```suggestion
  NOTIFY pgrst, 'reload schema'
```

schema_cache.rst
```diff
@@ -106,26 +111,52 @@ To refresh the cache without restarting the PostgREST server, send the server pr
...
+
+.. note::
+
+  The :ref:`db-channel-enabled` config option enables the notification channel by default.
+  This setting is incompatible with connection poolers such as PgBouncer in transaction pooling mode and should be set to ``false`` in that case.
```
steve-chavez, line 132:
```suggestion

The ``"pgrst"`` notification channel is enabled by default. For configuring the channel, see :ref:`db-channel` and :ref:`db-channel-enabled`.
```

schema_cache.rst
```diff
@@ -89,7 +89,12 @@ See the section :ref:`schema_reloading` to solve this issue.
 Schema Cache Reloading
 ----------------------
 
-To refresh the cache without restarting the PostgREST server, send the server process a SIGUSR1 signal:
+There are two ways of refreshing the cache without restarting the PostgREST server: using Unix signals or sending a database notification.
+
+Unix Signals
+~~~~~~~~~~~~
+
+Send a SIGUSR1 signal to the server process.
```
steve-chavez, line 97: My bad about this, I've now realized that this division doesn't make sense because SIGUSR2 would go in another page. So let's go back to the previous paragraph.

```suggestion
To refresh the cache without restarting the PostgREST server, send the server process a SIGUSR1 signal.
```

schema_cache.rst
```diff
@@ -106,26 +111,52 @@ To refresh the cache without restarting the PostgREST server, send the server pr
      # or in docker-compose
      docker-compose kill -s SIGUSR1 <service>
 
-The above is the manual way to do it. To automate cache reloads, use a database trigger like this:
+.. _schema_reloading_notify:
+
+NOTIFY
+~~~~~~
```
steve-chavez, line 117:
```suggestion
Reloading with NOTIFY
~~~~~~~~~~~~~~~~~~
```

schema_cache.rst
```diff
@@ -106,26 +111,52 @@ To refresh the cache without restarting the PostgREST server, send the server pr
...
+
+.. note::
+
+  The :ref:`db-channel-enabled` config option enables the notification channel by default.
+  This setting is incompatible with connection poolers such as PgBouncer in transaction pooling mode and should be set to ``false`` in that case.
+
+Automatic schema cache reloading
+********************************
+
+If the notification event is set to fire on a database event trigger, then automatic schema cache reloading is possible. For example:
```
steve-chavez, line 137:
```suggestion
You can do automatic schema cache reloading in a pure SQL way with an `event trigger <https://www.postgresql.org/docs/current/event-trigger-definition.html>`_ and ``NOTIFY``.
```

schema_cache.rst
laurenceisla: I think we could add that NOTIFY can help on Windows systems too,  according to #114. 

```suggestion
There are environments where you can't send the SIGUSR1 Unix Signal (like on managed containers in cloud services or on Windows systems). For this reason, PostgREST also allows you to reload its schema cache through PostgreSQL `NOTIFY <https://www.postgresql.org/docs/current/sql-notify.html>`_ as follows:
```

steve-chavez: Yeah! Mentioning Windows would be great too.

steve-chavez (approved):

Issue #409, opened 2021-06-22
Add doc for reloading configuration with SIGUSR2 and NOTIFY


steve-chavez:
This should be added in: https://postgrest.org/en/stable/configuration.html

- SIGUSR2 was added here: https://github.com/PostgREST/postgrest/pull/1544
- NOTIFY for config was added here: https://github.com/PostgREST/postgrest/pull/1729#issuecomment-757484308

PR #419, opened 2021-07-24
Add config reloading with NOTIFY and SIGUSR2

laurenceisla:
Added documentation according to #409 (only reloading with SIGUSR2 is included).

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -60,7 +60,7 @@ raw-media-types          String
 
 .. _config_reloading:
 
-Configuration reloading
+Configuration Reloading
 -----------------------
 
 To reload the configuration without restarting the PostgREST server send a SIGUSR2 signal to the server process.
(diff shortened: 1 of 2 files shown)


configuration.rst
```diff
@@ -58,6 +58,65 @@ role-claim-key           String  .role
...
+     docker-compose kill -s SIGUSR2 <service>
+
+You can also reload both the configuration and the schema cache by sending a SIGUSR1 signal.
+
+.. _config_reloading_notify:
+
+NOTIFY
+~~~~~~
+
+Send a `database notification <https://www.postgresql.org/docs/current/sql-notify.html>`_ from any PostgreSQL client by executing the ``NOTIFY`` command as follows:
```
steve-chavez, line 109: I think documenting NOTIFY config doesn't make much sense unless we also document the in-db configuration(https://github.com/PostgREST/postgrest/pull/1729). We could leave this part for another PR.

steve-chavez: The config reloading and SIGUSR2 docs are already looking good here.

steve-chavez (approved):
LGTM!

PR #423, opened 2021-07-31
Mention environment variables in the configuration section

mrexodia:
I had a `docker-compose.yml` with some environment variables in it and I couldn't really find any mention of them when searching Google for `PGRST_DB_URI` (other than the example). I think this small addition will make that a bit more discoverable.


laurenceisla:
@mrexodia Thanks for the help!

PR #425, opened 2021-08-12
Add partitioned tables in the schema cache

laurenceisla:
WIP

- [x] Changelog
- [x] Embedding documentation

### DIFF
--- a/api.rst
+++ b/api.rst
@@ -651,6 +651,49 @@ Embedded resources can be aliased and filters can be applied on these aliases:
 
   GET /films?select=*,90_comps:competitions(name),91_comps:competitions(name)&90_comps.year=eq.1990&91_comps.year=eq.1991 HTTP/1.1
 
+.. _embedding_partitioned_tables:
+
+Embedding Partitioned Tables
+----------------------------
+
+Embedding can also be done on partitioned tables referencing other resources.
+
+For example, let's create the ``box_office`` partitioned table to get the gross daily revenue of a film:
+
+.. code-block:: postgres
+
+  CREATE TABLE test.box_office (
+    bo_date DATE NOT NULL,
+    film_id INT REFERENCES test.films NOT NULL,
+    gross_revenue DECIMAL(12,2) NOT NULL,
+    PRIMARY KEY (bo_date, film_id)
+  ) PARTITION BY RANGE (bo_date);
+
+  -- Let's also create partitions for each month of 2021
+
+  CREATE TABLE box_office_2021_01 PARTITION OF test.box_office
+  FOR VALUES FROM ('2021-01-01') TO ('2021-01-31');
+
+  CREATE TABLE box_office_2021_02 PARTITION OF test.box_office
+  FOR VALUES FROM ('2021-02-01') TO ('2021-02-28');
+
+  -- and so until december 2021
+
+Since it contains the ``films_id`` foreign key, it is possible to embed ``box_office`` and ``films``:
+
+.. code-block:: http
+
+  GET /box_office?select=bo_date,gross_revenue,films(title)&gross_revenue=gte.1000000 HTTP/1.1
+
+Embedding is also possible between ``box_office`` partitions and the ``films`` table:
+
+.. code-block:: http
+
+  GET /films?select=title,box_office(bo_date,gross_revenue)&id=eq.10 HTTP/1.1
+
+.. note::
+  Partitioned tables can reference other tables since PostgreSQL 11, but can only be referenced from any other table since PostgreSQL 12.
+
 .. _embedding_views:
 
 Embedding Views
--- a/api.rst
+++ b/api.rst
@@ -656,9 +656,9 @@ Embedded resources can be aliased and filters can be applied on these aliases:
 Embedding Partitioned Tables
 ----------------------------
 
-Embedding can also be done on partitioned tables referencing other resources.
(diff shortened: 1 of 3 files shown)


api.rst
```diff
@@ -651,6 +651,49 @@ Embedded resources can be aliased and filters can be applied on these aliases:
...
+Embedding Partitioned Tables
+----------------------------
+
+Embedding can also be done between partitioned tables and other resources.
+
+For example, let's create the ``box_office`` partitioned table that has the gross daily revenue of a film:
+
+.. code-block:: postgres
+
+  CREATE TABLE test.box_office (
```
steve-chavez, line 665: Removing the test schema
```suggestion
  CREATE TABLE box_office (
```

api.rst
```diff
@@ -651,6 +651,49 @@ Embedded resources can be aliased and filters can be applied on these aliases:
 
   GET /films?select=*,90_comps:competitions(name),91_comps:competitions(name)&90_comps.year=eq.1990&91_comps.year=eq.1991 HTTP/1.1
 
+.. _embedding_partitioned_tables:
+
+Embedding Partitioned Tables
+----------------------------
+
+Embedding can also be done between partitioned tables and other resources.
```
steve-chavez, line 659:
```suggestion
Embedding can also be done between partitioned tables and other tables.
```

releases/upcoming.rst
```diff
@@ -0,0 +1,14 @@
...
+
+Upcoming
+========
+
+These are changes yet unreleased. If you'd like to try them out before a new official release, you can use a `nightly release <https://github.com/PostgREST/postgrest/releases/tag/nightly>`_.
+
+Added
+-----
+
+* Allow embedding, UPSERT, INSERT with Location response, OPTIONS request and OpenAPI support for partitioned tables.
```
steve-chavez, line 13: Needs a reference to "embedding_partitioned_tables"

releases/upcoming.rst
```diff
@@ -0,0 +1,14 @@
+.. |br| raw:: html
+
+   <br />
+
+Upcoming
+========
+
+These are changes yet unreleased. If you'd like to try them out before a new official release, you can use a `nightly release <https://github.com/PostgREST/postgrest/releases/tag/nightly>`_.
```
steve-chavez, line 8: Hm, nightlies are different now. Let's remove this for now.
```suggestion
```

api.rst
```diff
@@ -651,6 +651,49 @@ Embedded resources can be aliased and filters can be applied on these aliases:
...
+  GET /box_office?select=bo_date,gross_revenue,films(title)&gross_revenue=gte.1000000 HTTP/1.1
+
+Embedding is also possible between ``box_office`` partitions and the ``films`` table:
+
+.. code-block:: http
+
+  GET /films?select=title,box_office_2021_02(bo_date,gross_revenue)&rating=gt.8 HTTP/1.1
+
+.. note::
+  Partitioned tables can reference other tables since PostgreSQL 11, but can only be referenced from any other table since PostgreSQL 12.
```
steve-chavez, line 695:
```suggestion
  Partitioned tables can reference other tables since PostgreSQL 11 but can only be referenced from any other table since PostgreSQL 12.
```

steve-chavez:
@laurenceisla http://www.egull.co keeps failing linkcheck, maybe just comment the line for now.

api.rst
steve-chavez: Also, I think it would be nice to link to partition docs here: https://www.postgresql.org/docs/current/ddl-partitioning.html

steve-chavez (approved):

releases/upcoming.rst
steve-chavez: @monacoremo So as we discussed in https://github.com/PostgREST/postgrest/pull/1901#issuecomment-892939818, for trying the change here we'd need to link to [postgrest/actions/runs/1128881855](https://github.com/PostgREST/postgrest/actions/runs/1128881855) and tell users to download the artifacts right?

Some questions here:

- Do you know the artifacts' duration? I assume Github assigns them a lifetime.
- Try visiting the [run page](https://github.com/PostgREST/postgrest/actions/runs/1128881855) as incognito(not logged to GitHub), can you download the artifacts? I can't :confused:, also I wonder if logged-in non-team members can download them. Do you know if there's another way to link to the artifacts?

monacoremo: You need a Github account with read permissions on the repository (so any account as we are public, and anyone on Gitter), the artifacts are retained for 90 days: https://docs.github.com/en/actions/managing-workflow-runs/downloading-workflow-artifacts

For 'power users' to try a specific change right away (e.g. for whoever submitted the issue that got fixed), we should instruct them to go via the check on the relevant commit.

In order to have something to link to long term, we should tag a new version. In this case this might warrant a proper 8.0.1 release even. Otherwise I suggest 9.0.0-a1, which will create a prerelease, to be followed up with -a2 etc. The commit and date will be available on the release page. Or without the 1, just 9.0.0-a, if we want to overwrite the pre-release with new tags as development progresses. 

So for the instructions here I suggest:

 * To test unreleased changes, use one of the latest pre-releases on the release page (link)
 * If you need absolute bleeding edge (e.g. because you want to test if a specific commit fixed your issue), use the artifacts from the respective or latest CI run: https://github.com/PostgREST/postgrest/actions/workflows/ci.yaml?query=branch%3Amain (filtered for main branch)

PR #430, opened 2021-08-24
Add 'PostgREST' error code

laurenceisla:
Should be merged after https://github.com/PostgREST/postgrest/pull/1926 is approved.


releases/upcoming.rst
```diff
@@ -10,3 +10,6 @@ Added
 
 * Allow :ref:`embedding <embedding_partitioned_tables>`, UPSERT, INSERT with Location response, OPTIONS request and OpenAPI support for partitioned tables.
   |br| -- `@laurenceisla <https://github.com/laurenceisla>`_
+
+* Add ``"code": "PostgREST"`` to the error response body in order to differentiate those errors from the ones coming directly from the database. See :ref:`error_source`.
```
steve-chavez, line 14:
```suggestion
* Add ``"code": "PostgREST"`` to the error response body in order to differentiate PostgREST errors from errors coming in directly from the database. See :ref:`error_source`.
```

admin.rst
```diff
@@ -141,6 +141,40 @@ Server Version
 
 When debugging a problem it's important to verify the PostgREST version. At any time you can make a request to the running server and determine exactly which version is deployed. Look for the :code:`Server` HTTP response header, which contains the version number.
 
+.. _error_source:
+
+Error Source
+------------
+
+For the most part, error messages will come directly from the database with the usual code and message that PostgreSQL uses. PostgREST will then convert the error to JSON format with an HTTP status code (see the full list :ref:`here <status_codes>`). For instance, this is the error you will get when querying a nonexistent table:
```
steve-chavez, line 149: How about if we mention that the json response corresponds to pg error message format (MESSAGE, DETAIL, HINT
, ERRCODE): https://www.postgresql.org/docs/12/plpgsql-errors-and-messages.html

Actually we don't use ERRCODE, but we can emphasize that we stick to our structure for all error bodies and that `details` and `hint` can be null.

steve-chavez: The above just made me notice something.

Do PostgreSQL errors always come with `hint` and `details` as `null`?

Because in our case, when we don't have `hint`/`details` we simply omit the keys instead of having them as null, right?

Perhaps we should change that for consistency?

laurenceisla: >  * We'll use five-character error codes to keep some consistency with PostgreSQL (`"code": "PostgREST"` is now `"code": "PGRST"`)
> 
>  * Other errors coming from Hasql will have their own codes e.g. `PGCON` for `ConnectionError`

This is no longer valid, we continued the discussion on the PR that implements this change: https://github.com/PostgREST/postgrest/pull/1926#issuecomment-911695893

docs/admin.rst
```diff
@@ -370,6 +405,288 @@ Nginx rewrite rules allow you to simulate the familiar URL convention. The follo
...
+| PGRST001      | PostgreSQL service not running.             |
++---------------+---------------------------------------------+
+| .. _pgrstx02: | Could not connect with the database due to  |
+|               | an internal error.                          |
+| PGRST002      |                                             |
++---------------+---------------------------------------------+
+| .. _pgrstx03: | Could not connect with the database when    |
+|               | building the :ref:`schema_cache` due to the |
+| PGRST003      | PostgreSQL service not running.             |
++---------------+---------------------------------------------+
```
laurenceisla, line 436: Added this as an alternative to using subheadings. Opinions on which one would be better to use are welcome.

steve-chavez: @laurenceisla Yeah, this looks much better!

Seems the anchor link is not clickable though, any way to work around that?

Even if that doesn't work I'd still favor it over subheadings.

---

Found a discussion about anchor links in table rows https://github.com/STEllAR-GROUP/hpx/issues/4495. Seems it's possible to do them.

docs/admin.rst
```diff
@@ -153,6 +153,41 @@ Server Version
 
 When debugging a problem it's important to verify the PostgREST version. At any time you can make a request to the running server and determine exactly which version is deployed. Look for the :code:`Server` HTTP response header, which contains the version number.
 
+.. _error_source:
+
+Error Source
+------------
+
+For the most part, error messages will come directly from the database with the same `format that PostgreSQL uses <https://www.postgresql.org/docs/current/plpgsql-errors-and-messages.html>`_, in other words, PostgREST will convert the ``MESSAGE``, ``DETAIL``, ``HINT`` and ``ERRCODE`` from the PostgreSQL error to JSON format and add an HTTP status code to the response (see :ref:`status_codes`). For instance, this is the error you will get when querying a nonexistent table:
```
steve-chavez, line 161: How about moving this whole doc to a dedicated "Errors" reference page?

We could change this section https://postgrest--430.org.readthedocs.build/en/430/api.html#http-status-codes to be there too.

steve-chavez: Looks great now!

docs/admin.rst
laurenceisla: > Found a discussion about anchor links in table rows https://github.com/STEllAR-GROUP/hpx/issues/4495. Seems it's possible to do them.

That's good, although it keeps the anchor link visible at all times, not at hover only. I'll leave that solution (the hover anchor links) for another PR.

steve-chavez (approved):

PR #433, opened 2021-09-02
Add link and explanation to download the latest unreleased builds

laurenceisla:
Now that Nightlies are deprecated, the latest builds can be downloaded from the resulting Artifacts from CI runs.


steve-chavez:
Not related to the PR, but linkcheck is failing because https://github.com/datrium/postgrest-pyclient no longer exists - maybe remove it.

PR #436, opened 2021-09-22
Update install.rst

drTr0jan:
Updated FreeBSD dependencies as in `bsd.default-versions.mk`

### DIFF
--- a/install.rst
+++ b/install.rst
@@ -213,7 +213,7 @@ You can build PostgREST from source with `Stack <https://github.com/commercialha
   =====================  =======================================
   Ubuntu/Debian          libpq-dev, libgmp-dev, zlib1g-dev
   CentOS/Fedora/Red Hat  postgresql-devel, zlib-devel, gmp-devel
-  BSD                    postgresql95-client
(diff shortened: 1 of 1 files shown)


install.rst
```diff
@@ -213,7 +213,7 @@ You can build PostgREST from source with `Stack <https://github.com/commercialha
   =====================  =======================================
   Ubuntu/Debian          libpq-dev, libgmp-dev, zlib1g-dev
   CentOS/Fedora/Red Hat  postgresql-devel, zlib-devel, gmp-devel
-  BSD                    postgresql95-client
+  BSD                    postgresql12-client
```
wolfgangwalther, line 216: Why not `postgresql13-client`, which is the latest released version?

(note, I have no experience with BSD, so maybe I'm missing something obvious)

wolfgangwalther: Ah, you already said that... because of:
https://cgit.freebsd.org/ports/tree/Mk/bsd.default-versions.mk#n109

I see...

steve-chavez (approved):

PR #437, opened 2021-09-30
Specify external connection pooler

laurenceisla:
Specify that the connection poolers are external and not from PostgREST itself.

### DIFF
--- a/admin.rst
+++ b/admin.rst
@@ -125,7 +125,7 @@ Nginx rate limiting is general and indiscriminate. To rate limit each authentica
 Using External Connection Poolers
 ---------------------------------
 
-In order to increase performance, PostgREST uses prepared statements by default. However, this setting is incompatible with connection poolers such as PgBouncer working in transaction pooling mode. In this case, you need to set the :ref:`db-prepared-statements` config option to ``false``. On the other hand, session pooling is fully compatible with PostgREST, while statement pooling is not compatible at all.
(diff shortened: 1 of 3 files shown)


admin.rst
```diff
@@ -120,10 +120,10 @@ The burst argument tells Nginx to start dropping requests if more than five queu
...
 
-.. _connection_poolers:
+.. _external_connection_poolers:
 
-Using Connection Poolers
-------------------------
+Using External Connection Poolers
+---------------------------------
 
 In order to increase performance, PostgREST uses prepared statements by default. However, this setting is incompatible with connection poolers such as PgBouncer working in transaction pooling mode. In this case, you need to set the :ref:`db-prepared-statements` config option to ``false``. On the other hand, session pooling is fully compatible with PostgREST, while statement pooling is not compatible at all.
```
steve-chavez, line 128: We need to clarify here that PostgREST already has its own pool, maybe linking to https://postgrest.org/en/v8.0/configuration.html#db-pool.

Also maybe mention "external":
```suggestion
In order to increase performance, PostgREST uses prepared statements by default. However, this setting is incompatible with external connection poolers such as PgBouncer working in transaction pooling mode. In this case, you need to set the :ref:`db-prepared-statements` config option to ``false``. On the other hand, session pooling is fully compatible with PostgREST, while statement pooling is not compatible at all.
```

steve-chavez (approved):

PR #440, opened 2021-10-07
Add compatibility to PostgreSQL 14

laurenceisla:
Related to https://github.com/PostgREST/postgrest/pull/1968

### DIFF
--- a/api.rst
+++ b/api.rst
@@ -1588,8 +1588,8 @@ You can access request headers, cookies and JWT claims by reading GUC variables 
 
 .. code-block:: postgresql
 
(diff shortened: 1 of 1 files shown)


steve-chavez:
@laurenceisla Do you think we could add a new section(subheading) to the configuration page regarding "backward compatibility config parameters"?

Something like: https://techdocs.broadcom.com/us/en/symantec-security-software/identity-security/siteminder/12-8/configuring/web-agent-configuration/advanced-configuration-settings/backward-compatibility-settings.html

This way new users(on pg14) can ignore this new parameter - it can be confusing why is the option present otherwise.

PR #451, opened 2021-10-26
Allow rpc POST with single unnamed parameter

laurenceisla:
Documentation for https://github.com/PostgREST/postgrest/pull/1927

Replaces https://github.com/PostgREST/postgrest-docs/pull/431

### DIFF
--- a/api.rst
+++ b/api.rst
@@ -1877,9 +1877,9 @@ To send raw JSON, you can avoid using the ``Prefer: params=single-object`` heade
 
 .. note::
 
-  If an overloaded function has a single ``json`` or ``jsonb`` unnamed parameter, PostgREST will call this function as a fallback provided that no other overloaded function is found with the parameters sent in the request.
+  If an overloaded function has a single ``json`` or ``jsonb`` unnamed parameter, PostgREST will call this function as a fallback provided that no other overloaded function is found with the parameters sent in the POST request.
 
-To upload raw binary, the parameter type must be ``bytea`` and the header ``Content-Type: application/octet-stream`` must be included in the request. Similarly, to upload raw text, the parameter type must be ``text`` and the header ``Content-Type: text/plain`` must be included in the request.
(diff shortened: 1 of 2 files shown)


api.rst
```diff
@@ -1316,6 +1318,54 @@ You can also call a function that takes a single parameter of type JSON by sendi
...
+
+  { "x": 4, "y": 2 }
+
+  8
+
+.. _s_proc_single_unnamed_json_overloaded:
+
+.. important::
+
+  Functions with a single unnamed JSON parameter must **not** be overloaded, otherwise PostgREST will not be able to resolve a POST request to the function and will send this response:
```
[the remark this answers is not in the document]
laurenceisla: After https://github.com/PostgREST/postgrest/pull/1996 is merged, this warning is not needed anymore.

steve-chavez (approved):

api.rst
```diff
@@ -1839,6 +1841,42 @@ You can also call a function that takes a single parameter of type JSON by sendi
...
   8
 
+.. _s_proc_single_unnamed:
+
+Calling functions with a single unnamed parameter
+-------------------------------------------------
+
+You can make a POST request to a function with a single unnamed parameter to send raw ``json/jsonb``, ``bytea`` or ``text`` data.
+
+To send raw ``json/jsonb``, you can avoid using the ``Prefer: params=single-object`` header if the function has a single unnamed JSON parameter and the header ``Content-Type: application/json`` is included in the request.
```
wolfgangwalther, line 1851: Should we deprecate `params=single-object` or is there still a valid use-case for it?

Also, I think the fallback behaviour for the overloaded json case should be mentioned, WDYT?

steve-chavez: > Should we deprecate params=single-object or is there still a valid use-case for it?

The unnamed json parameter doesn't work for GET, while `params=single-object` does.

Also, right now unnamed json works for webhooks that only send a JSON body. There are other webhooks that also send some query params besides the JSON body and this makes the function call fail(maybe the solution is to add a second unnamed json param for "absorbing" the querystring parameters).

In any case, I think it's too early to deprecate(next version could), we need the new unnamed json feature to be proven first.

wolfgangwalther: Makes a ton of sense!

PR #458, opened 2021-11-12
Add the logical operators section in horizontal filtering

laurenceisla:
The `and` and `or` operators are somewhat "hidden" in the Horizontal Filtering section. This PR groups them in a subheading and adds them to the operator list.

### DIFF
--- a/api.rst
+++ b/api.rst
@@ -51,30 +51,6 @@ Multiple parameters can be logically conjoined by:
 
     curl "http://localhost:3000/people?age=gte.18&student=is.true"
 
-Multiple parameters can be logically disjoined by:
-
-.. tabs::
-
-  .. code-tab:: http
-
-    GET /people?or=(age.gte.14,age.lte.18) HTTP/1.1
-
-  .. code-tab:: bash Curl
-
-    curl "http://localhost:3000/people?or=(age.gte.14,age.lte.18)"
-
-Complex logic can also be applied:
-
-.. tabs::
-
-  .. code-tab:: http
-
-    GET /people?and=(grade.gte.90,student.is.true,or(age.gte.14,age.is.null)) HTTP/1.1
-
-  .. code-tab:: bash Curl
-
-    curl "http://localhost:3000/people?and=(grade.gte.90,student.is.true,or(age.gte.14,age.is.null))"
-
 .. _operators:
 
 Operators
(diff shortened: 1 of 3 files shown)


api.rst
```diff
@@ -139,6 +115,49 @@ The view will provide a new endpoint:
 
     curl "http://localhost:3000/fresh_stories"
 
+.. _logical_operators:
+
+Logical operators
+~~~~~~~~~~~~~~~~~
+
+Multiple parameters can be logically **conjoined** using ``and``:
```
steve-chavez, line 123: This paragraph makes me think that to combine filters with `and` I have to use the `and=(..)` notation.

So I think this section should start with something like "by default filters are combined with AND but you can use OR with the following notation..". OR should be the main focus.

api.rst
```diff
@@ -111,11 +87,11 @@ sr            :code:`>>`                strictly right of
 nxr           :code:`&<`                does not extend to the right of, e.g. :code:`?range=nxr.(1,10)`
 nxl           :code:`&>`                does not extend to the left of
 adj           :code:`-|-`               is adjacent to, e.g. :code:`?range=adj.(1,10)`
-not           :code:`NOT`               negates another operator, see below
+and           :code:`AND`               logical :code:`AND`, see :ref:`logical_operators`
```
steve-chavez, line 90: As I mentioned on the above comment, AND is not really needed because it's the default. I can't find the link now, but I read somewhere that our syntax was complicated and error prone because it was thought that `and=(..)` had to be used for boolean logic.

IIRC, the `and` operator was implemented for completion, `or` was the main feature.

Perhaps we should remove it from here and just mention it on the Logical operators section.

wolfgangwalther: I disagree - it's a valid operator so we should list it here for completeness.

Maybe we can just add a hint about that default behaviour?

steve-chavez: > it's a valid operator so we should list it here for completeness.

Hm, yeah. Maybe we can list the `AND` as the last row on the table since its usage will be rare.

> Maybe we can just add a hint about that default behaviour?

Sounds good.

api.rst
```diff
@@ -139,6 +115,37 @@ The view will provide a new endpoint:
...
+
+.. tabs::
+
+  .. code-tab:: http
+
+    GET /people?people?grade=gte.90&student=is.true&or=(age.eq.14,not.and=(age.gte.11,age.lte.17)) HTTP/1.1
+
+  .. code-tab:: bash Curl
+
+    curl "http://localhost:3000/people?grade=gte.90&student=is.true&or=(age.eq.14,not.and(age.gte.11,age.lte.17))"
```
steve-chavez, line 147: The http example needs correction, has invalid syntax

steve-chavez (approved):

PR #459, opened 2021-11-16
Remove partitions from the schema cache

laurenceisla:
Related to https://github.com/PostgREST/postgrest/pull/2022

### DIFF
--- a/api.rst
+++ b/api.rst
@@ -1084,18 +1084,6 @@ Since it contains the ``films_id`` foreign key, it is possible to embed ``box_of
 
     curl "http://localhost:3000/box_office?select=bo_date,gross_revenue,films(title)&gross_revenue=gte.1000000"
 
-Embedding is also possible between ``box_office`` partitions and the ``films`` table:
-
-.. tabs::
-
-  .. code-tab:: http
-
-    GET /films?select=title,box_office_2021_02(bo_date,gross_revenue)&rating=gt.8 HTTP/1.1
-
-  .. code-tab:: bash Curl
-
-    curl "http://localhost:3000/films?select=title,box_office_2021_02(bo_date,gross_revenue)&rating=gt.8"
-
 .. note::
   Partitioned tables can reference other tables since PostgreSQL 11 but can only be referenced from any other table since PostgreSQL 12.
 
--- a/api.rst
+++ b/api.rst
@@ -1085,7 +1085,8 @@ Since it contains the ``films_id`` foreign key, it is possible to embed ``box_of
     curl "http://localhost:3000/box_office?select=bo_date,gross_revenue,films(title)&gross_revenue=gte.1000000"
 
 .. note::
-  Partitioned tables can reference other tables since PostgreSQL 11 but can only be referenced from any other table since PostgreSQL 12.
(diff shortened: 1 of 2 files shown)


steve-chavez:
Should we mention somewhere(maybe in a note) that partitions are ignored because they cause conflicts when embedding?

Some of the reasoning in https://github.com/PostgREST/postgrest/issues/1783#issuecomment-961121977 could also be included.

laurenceisla:
> Should we mention somewhere(maybe in a note) that partitions are ignored because they cause conflicts when embedding?
> 
> Some of the reasoning in [PostgREST/postgrest#1783 (comment)](https://github.com/PostgREST/postgrest/issues/1783#issuecomment-961121977) could also be included.

Yes, I think both should be mentioned, but mostly around the fact that partitions shouldn't be exposed and the embedding conflict solution as a consequence of that.

api.rst
```diff
@@ -1084,20 +1084,9 @@ Since it contains the ``films_id`` foreign key, it is possible to embed ``box_of
...
-    GET /films?select=title,box_office_2021_02(bo_date,gross_revenue)&rating=gt.8 HTTP/1.1
-
-  .. code-tab:: bash Curl
-
-    curl "http://localhost:3000/films?select=title,box_office_2021_02(bo_date,gross_revenue)&rating=gt.8"
-
 .. note::
-  Partitioned tables can reference other tables since PostgreSQL 11 but can only be referenced from any other table since PostgreSQL 12.
+  * Partitioned tables can reference other tables since PostgreSQL 11 but can only be referenced from any other table since PostgreSQL 12.
+  * Partitions are not included in the :ref:`schema_cache` since they should not be exposed to such features. This also avoids embedding ambiguity for the parent partitioned table.
```
steve-chavez, line 1089:
```suggestion
  * Embedding between the partitioned table and its partitions is not allowed because it leads to ambiguity errors(see :ref:`embed_disamb`). :ref:`custom_queries` can be used if this is needed.
```

laurenceisla: Maybe it should be changed to something like "Embedding on partitions is not allowed because it leads to ambiguity errors between them and their parent partitioned table...". The ambiguity arises mostly when other table embeds the parent table.

api.rst
```diff
@@ -1086,7 +1086,7 @@ Since it contains the ``films_id`` foreign key, it is possible to embed ``box_of
 
 .. note::
   * Partitioned tables can reference other tables since PostgreSQL 11 but can only be referenced from any other table since PostgreSQL 12.
-  * Partitions are not included in the :ref:`schema_cache` since they should not be exposed to such features. This also avoids embedding ambiguity for the parent partitioned table.
+  * Embedding on partitions is not allowed because it leads to ambiguity errors between them and their parent partitioned table (see :ref:`embed_disamb`). :ref:`custom_queries` can be used if this is needed.
```
steve-chavez, line 1089:
```suggestion
  * Embedding on partitions is not allowed because it leads to ambiguity errors (see :ref:`embed_disamb`) between them and their parent partitioned table. :ref:`custom_queries` can be used if this is needed.
```

steve-chavez (approved):

PR #483, opened 2022-01-07
Update configuration page

wolfgangwalther:
This is kind of a rework of the configuration page, including:
- Adding the missing `db-root-spec` config parameter
- Renaming the config parameters according to https://github.com/PostgREST/postgrest/pull/1679
- Sorting all config parameters alphabetically
- Adding names for environment variables and database settings for each parameter
- Re-organizing the whole page, so that the list of parameters is last (so that the environment and database sections come before the list, which already mentions the names for those settings)

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -24,377 +24,13 @@ The configuration file must contain a set of key value pairs. At minimum you mus
   db-uri       = "postgres://user:pass@host:5432/dbname"
 
   # The name of which database schema to expose to REST clients
-  db-schema    = "api"
+  db-schemas   = "api"
 
   # The database role to use when no client authentication is provided.
   # Can (and should) differ from user in db-uri
   db-anon-role = "anon"
 
 The user specified in the db-uri is also known as the authenticator role. For more information about the anonymous vs authenticator roles see the :ref:`roles`.
-
-.. _config_full_list:
-
-Here is the full list of configuration parameters.
-
-======================== ======= ================= ========
-Name                     Type    Default           Required
-======================== ======= ================= ========
-db-uri                   String                    Y
-db-schema                String                    Y
-db-anon-role             String                    Y
-db-pool                  Int     10
-db-pool-timeout          Int     10
-db-extra-search-path     String  public
-db-channel               String  pgrst
-db-channel-enabled       Boolean True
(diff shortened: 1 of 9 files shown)


configuration.rst
```diff
@@ -24,159 +24,334 @@ The configuration file must contain a set of key value pairs. At minimum you mus
...
+
+  A schema-qualified stored procedure to call instead of the built-in OpenAPI code on `/`. This provides the opportunity to implement a custom response on the root endpoint.
+
+.. _db-schemas:
+
+db-schemas
 ----------
 
-  The name of the notification channel that PostgREST uses for :ref:`schema_reloading` and configuration reloading.
+  *For backwards compatibility, this config parameter is also available in singular as "db-schema".*
```
wolfgangwalther, line 313: As noticed in https://github.com/PostgREST/postgrest-docs/issues/487#issuecomment-1007938516, in-database configuration does not support the old aliases. Maybe we should add a hint here.

wolfgangwalther:
> > Adding the missing db-root-spec config parameter
> 
> Hm, except this one. 
> [...]
> How about not including in the docs?

Ok, cool.

configuration.rst
wolfgangwalther: Went without a hint for now. I think having the new names used everywhere should be enough to make people use it.

wolfgangwalther:
> The `db-root-spec` idea is fraught with performance deficiencies, in fact the whole json generation for OpenAPI inside PostgREST is. I mentioned this before on [PostgREST/postgrest#1794 (comment)](https://github.com/PostgREST/postgrest/pull/1794#issuecomment-846364797) and lately I've been finding instances where even a bigger schema results in postgrest consuming 100% CPU and not finishing while trying to respond to the root endpoint.
> 
> Seems `postgrest-contrib` will also be needed for performance and not just extensibility.
> 
> So it's likely that `db-root-spec` will be removed and it doesn't seem right to encourage users into relying on it, maybe not even in an experimental way.

I understand the performance problem is related to the spec GUC (and the big OpenAPI query in core) right now - and not to the `db-root-spec` concept overall. So I guess `db-root-spec` will stay - although maybe the requirements for the function to be called might change. So it's good to keep it undocumented for now.

Issue #445, opened 2021-10-19
Add more fine grained event trigger for automatic schema cache reloading


steve-chavez:
Right now https://postgrest.org/en/v8.0/schema_cache.html#automatic-schema-cache-reloading reloads on every DDL change but we don't need to reload on changes on triggers, grants, policies, etc.

See https://github.com/PostgREST/postgrest/issues/1512#issuecomment-643687504 for a basic idea on how to do a more fine grained even trigger.

steve-chavez:
A problem with the current event trigger is that a `create temp table` fires it.

Just reproduced it with a function like:

```sql
create or replace function what() returns text as $$ 
  create temp table if not exists wut(); 
  select 'what'::text 
$$ language sql;
```

Each call with POST will print:

```
127.0.0.1 - - [04/Jan/2022:21:18:18 -0500] "POST /rpc/what HTTP/1.1" 200 - "" "curl/7.76.1"
04/Jan/2022:21:18:18 -0500: Schema cache loaded

127.0.0.1 - - [04/Jan/2022:21:18:20 -0500] "POST /rpc/what HTTP/1.1" 200 - "" "curl/7.76.1"
04/Jan/2022:21:18:21 -0500: Schema cache loaded
```

A GET does fail:

```json
{"hint":null,"details":null,"code":"25006","message":"cannot execute CREATE TABLE in a read-only transaction"}
```

This of course affects the performance of functions that use temp tables for internal logic.

steve-chavez:
@wolfgangwalther Can you remind me why we need collations?

---

I've settled on the following function for now(also took [event trigger matrix](https://www.postgresql.org/docs/current/event-trigger-matrix.html) as reference), since there's no event for `CREATE TEMP TABLE` I excluded every object created on `pg_temp`.

```sql
CREATE OR REPLACE FUNCTION pgrst_watch() RETURNS event_trigger AS $$
DECLARE
  obj record;
BEGIN
  IF tg_tag IN (
    -- schemas
    'CREATE SCHEMA', 'ALTER SCHEMA', 'DROP SCHEMA'
    -- tables
  , 'CREATE TABLE', 'CREATE TABLE AS', 'SELECT INTO', 'ALTER TABLE', 'DROP TABLE'
    -- foreign tables
  , 'CREATE FOREIGN TABLE', 'ALTER FOREIGN TABLE', 'DROP FOREIGN TABLE'
    -- views
[...]

PR #489, opened 2022-01-12
Add finer-grained event trigger

steve-chavez:
Resolves https://github.com/PostgREST/postgrest-docs/issues/445.

### DIFF
--- a/schema_cache.rst
+++ b/schema_cache.rst
@@ -173,36 +173,24 @@ reloading when creating temporary tables(``CREATE TEMP TABLE``) inside functions
   -- watch create and alter
   CREATE OR REPLACE FUNCTION pgrst_ddl_watch() RETURNS event_trigger AS $$
   DECLARE
-    obj record;
+    cmd record;
   BEGIN
-    FOR obj IN SELECT * FROM pg_event_trigger_ddl_commands()
+    FOR cmd IN SELECT * FROM pg_event_trigger_ddl_commands()
     LOOP
-      IF obj.command_tag IN (
-        -- schemas
+      IF cmd.command_tag IN (
         'CREATE SCHEMA', 'ALTER SCHEMA'
-        -- tables
       , 'CREATE TABLE', 'CREATE TABLE AS', 'SELECT INTO', 'ALTER TABLE'
-        -- foreign tables
       , 'CREATE FOREIGN TABLE', 'ALTER FOREIGN TABLE'
-        -- views
       , 'CREATE VIEW', 'ALTER VIEW'
-        -- materialized views
       , 'CREATE MATERIALIZED VIEW', 'ALTER MATERIALIZED VIEW'
-        -- functions
       , 'CREATE FUNCTION', 'ALTER FUNCTION'
-        -- triggers
       , 'CREATE TRIGGER'
-        -- types
       , 'CREATE TYPE'
-        -- rules
       , 'CREATE RULE'
-        -- comments
       , 'COMMENT'
       )
       -- don't notify in case of CREATE TEMP table or other objects created on pg_temp
-      AND obj.schema_name is distinct from 'pg_temp'
-      -- don't notify on extensions changes
-      AND obj.in_extension IS FALSE
+      AND cmd.schema_name is distinct from 'pg_temp'
       THEN
         NOTIFY pgrst, 'reload schema';
       END IF;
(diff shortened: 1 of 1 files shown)


schema_cache.rst
```diff
@@ -160,4 +160,84 @@ To disable auto reloading, drop the trigger:
...
+      , 'CREATE TRIGGER'
+        -- types
+      , 'CREATE TYPE'
+        -- rules
+      , 'CREATE RULE'
+        -- comments
+      , 'COMMENT'
+      )
+      -- don't notify in case of CREATE TEMP table or other objects created on pg_temp
+      AND obj.schema_name is distinct from 'pg_temp'
```
steve-chavez, line 203: `is distinct` is used because in case of CREATE SCHEMA the schema name is NULL

schema_cache.rst
```diff
@@ -160,4 +160,84 @@ To disable auto reloading, drop the trigger:
...
+      -- don't notify on extensions changes
+      AND obj.in_extension IS FALSE
```
steve-chavez, line 205: Haven't tested how reliable this actually is. Also there's no equivalent for drops.

steve-chavez, line 205: No equivalent for DROPs on this one.

schema_cache.rst
```diff
@@ -160,4 +160,84 @@ To disable auto reloading, drop the trigger:
...
+      THEN
+        NOTIFY pgrst, 'reload schema';
+      END IF;
+    END LOOP;
+  END; $$ LANGUAGE plpgsql;
+
+  -- watch drop
+  CREATE OR REPLACE FUNCTION pgrst_drop_watch() RETURNS event_trigger AS $$
```
steve-chavez, line 213: Confirmed that:

- No need to catch DROP OWNED in the other trigger, as it generates a lot of individual DROPs that are caught here.
- The built-in debounce in NOTIFY works for these cases - only one NOTIFY is sent for many dropped objects.

schema_cache.rst
```diff
@@ -160,4 +160,84 @@ To disable auto reloading, drop the trigger:
...
+        -- functions
+      , 'CREATE FUNCTION', 'ALTER FUNCTION'
+        -- triggers
+      , 'CREATE TRIGGER'
+        -- types
+      , 'CREATE TYPE'
+        -- rules
+      , 'CREATE RULE'
+        -- comments
+      , 'COMMENT'
```
wolfgangwalther, line 200:
```suggestion
        'CREATE SCHEMA', 'ALTER SCHEMA'
      , 'CREATE TABLE', 'CREATE TABLE AS', 'SELECT INTO', 'ALTER TABLE'
      , 'CREATE FOREIGN TABLE', 'ALTER FOREIGN TABLE'
      , 'CREATE VIEW', 'ALTER VIEW'
      , 'CREATE MATERIALIZED VIEW', 'ALTER MATERIALIZED VIEW'
      , 'CREATE FUNCTION', 'ALTER FUNCTION'
      , 'CREATE TRIGGER'
      , 'CREATE TYPE'
      , 'CREATE RULE'
      , 'COMMENT'
```

Pretty obvious from the code itself what those lines are about ;)

schema_cache.rst
```diff
@@ -160,4 +160,84 @@ To disable auto reloading, drop the trigger:
...
+reloading when creating temporary tables(``CREATE TEMP TABLE``) inside functions.
+
+.. code-block:: postgresql
+
+  -- watch create and alter
+  CREATE OR REPLACE FUNCTION pgrst_ddl_watch() RETURNS event_trigger AS $$
+  DECLARE
+    obj record;
+  BEGIN
+    FOR obj IN SELECT * FROM pg_event_trigger_ddl_commands()
```
wolfgangwalther, line 178:
```suggestion
    cmd record;
  BEGIN
    FOR cmd IN SELECT * FROM pg_event_trigger_ddl_commands()
```

I think in this trigger `cmd` is better, because `pg_event_trigger_ddl_commands()` returns commands, not objects.

schema_cache.rst
wolfgangwalther: I don't understand the comment.

---

Apart from this:

Why are we not notifying on extension changes? If the extension's functions are in a public schema we are exposing them - so we should update?

What if I actually deploy my API code as an extension itself?

wolfgangwalther: Ah, ok - so the way it was before, we would have reloaded when extensions object were dropped, but not when they were created? Even better to have it consistent now.

PR #491, opened 2022-01-16
fix: remove net option from docker cmd

cemremengu:
net and port binding options cannot be used together

### DIFF
--- a/install.rst
+++ b/install.rst
@@ -129,7 +129,7 @@ The first way to run PostgREST in Docker is to connect it to an existing native 
 .. code-block:: bash
 
   # Run the server
-  docker run --rm --net=host -p 3000:3000 \
+  docker run --rm -p 3000:3000 \
     -e PGRST_DB_URI="postgres://postgres@localhost/postgres" \
     -e PGRST_DB_SCHEMA="public" \
     -e PGRST_DB_ANON_ROLE="postgres" \


wolfgangwalther:
Hm. This falls a bit short. There is a whole paragraph about macOS not supporting `--net` - which would be useless after the change.

But before we go ahead and rip that out, too: Will removing the `--net` even work? The example is about a dockerized postgrest with postgresql on the host machine. So there should still be some way to connect to the database - and `localhost` won't do it.

Wouldn't the solution be to remove the `port` instead, because when using `--net=host`, postgrest will listen on `localhost:3000` anyway?

cemremengu:
Docker ignores the `port` when `--net` is present i.e: `WARNING: Published ports are discarded when using host network mode`

However, I didn't know about macOS so in that case we can keep this as is.

and yes this is correct

> Wouldn't the solution be to remove the port instead, because when using --net=host, postgrest will listen on localhost:3000 anyway?

wolfgangwalther:
I'll reopen this, because I think we should in fact do *something* about it. Having that warning seems not useful. Not sure what exactly, we'll do, but this will keep it visible for now.

steve-chavez:
Maybe we need tabs(like we do for the curl/http snippets) for the docker installation methods.

wolfgangwalther:
Fixed this in #499.

PR #493, opened 2022-01-22
Run PostgREST with zero config. 

wolfgangwalther:
Related to https://github.com/PostgREST/postgrest/pull/2112

### DIFF
--- a/configuration.rst
+++ b/configuration.rst
@@ -39,31 +39,6 @@ Environment Variables
 
 You can also set these :ref:`configuration parameters <config_full_list>` using environment variables. They are capitalized, have a ``PGRST_`` prefix, and use underscores. For example: ``PGRST_DB_URI`` corresponds to ``db-uri`` and ``PGRST_APP_SETTINGS_*`` to ``app.settings.*``.
 
-.. _config_reloading:
-
-Configuration Reloading
-=======================
-
-To reload the configuration without restarting the PostgREST server, send a SIGUSR2 signal to the server process.
-
-.. code:: bash
-
-  killall -SIGUSR2 postgrest
-
-This method does not reload :ref:`env_variables_config` and it will not work for reloading a Docker container configuration. In these cases, you need to restart the PostgREST server or use the :ref:`in_db_config` as an alternative.
-
-.. important::
-
-  The following settings will not be reread when reloading the configuration. You will need to restart PostgREST in that case.
-
-    * :ref:`db-uri`
-    * :ref:`db-pool`
-    * :ref:`db-pool-timeout`
-    * :ref:`server-host`
-    * :ref:`server-port`
-    * :ref:`server-unix-socket`
-    * :ref:`server-unix-socket-mode`
-
 .. _in_db_config:
 
 In-Database Configuration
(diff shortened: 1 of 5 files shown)


configuration.rst
```diff
@@ -369,9 +373,7 @@ db-uri
 
   When running PostgREST on the same machine as PostgreSQL, it is also possible to connect to the database using a `Unix socket <https://en.wikipedia.org/wiki/Unix_domain_socket>`_ and the `Peer Authentication method <https://www.postgresql.org/docs/current/auth-peer.html>`_ as an alternative to TCP/IP communication and authentication with a password, this also grants higher performance.  To do this you can omit the host and the password, e.g. ``postgres://user@/dbname``, see the `libpq connection string <https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING>`_ documentation for more details.
 
-  On older systems like Centos 6, with older versions of libpq, a different db-uri syntax has to be used. In this case the URI is a string of space separated key-value pairs (key=value), so the example above would be :code:`"host=host user=user port=5432 dbname=dbname password=pass"`.
```
wolfgangwalther, line 372: I think this was relevant pre 9.2 - we should be able to remove that safely by now.

configuration.rst
```diff
@@ -3,7 +3,14 @@
 Configuration
 =============
 
-PostgREST reads a configuration file to determine information about the database and how to serve client requests. There is no predefined location for this file, you must specify the file path as the one and only argument to the server:
+Without configuration, PostgREST won't be able to serve any requests. At the minimum it needs either a role to serve anonymous requests with - or a secret to use for JWT authentication. To connect to a database it will read `libpq environment variables <https://www.postgresql.org/docs/current/libpq-envars.html>`_ by default. The user specified in ``PGUSER`` is also known as the authenticator role. For more information about the anonymous vs authenticator roles see the :ref:`roles`.
```
steve-chavez, line 6: Hm, I think it's a bit confusing mentioning the LIBPQ env vars right away. Perhaps we can move the PGUSER, authenticator and pg docs to another section? Could be inside "Environment variables" as "LIBPQ environment variables".

configuration.rst
```diff
@@ -13,24 +20,19 @@ PostgREST reads a configuration file to determine information about the database
...
-The configuration file must contain a set of key value pairs. At minimum you must include these keys:
+The configuration file must contain a set of key value pairs. At a minimum you should include one of the following keys:
 
 .. code::
 
   # postgrest.conf
 
-  # The standard connection URI format, documented at
-  # https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING
-  db-uri       = "postgres://user:pass@host:5432/dbname"
```
steve-chavez, line 24: This still seems like valuable information, specially for remote pg connections.

I think instead of replacing the old ways we should add a subsection like "default postgres connection".

configuration.rst
wolfgangwalther: I think the libpq environment variables are different from the other environment variables. Especially because of how they interact with the db-uri string. I'm think about adding a separate section for "database connection" in general, which would describe the different ways to do so.

configuration.rst
wolfgangwalther: > This still seems like valuable information

Right. Looking at the whole changes, I think I removed a bit too much without replacement. Maybe the "database connection" section mentioned above could instead pick that up.

> I think instead of replacing the old ways we should add a subsection like "default postgres connection".

Hm. I would like to describe the default first and then go from there. That seems more natural to me.

configuration.rst
wolfgangwalther: Didn't go with a separate section, but still expanded on that. WDYT? Open for suggestions.

wolfgangwalther: > Hm, I think it's a bit confusing mentioning the LIBPQ env vars right away.

I thought about that part - and I feel like the most important thing on the configuration reference page is to get the facts / behaviour exactly right. I didn't touch the tutorials too much on purpose: Those still provide the connection string via plain config file, because that's easy enough to understand.

But especially the interaction between connection string and libpq environment variables deserves notice. Even when using a dedicated connection string, libpq will still read those parameters that are not set from env variables. So that could create some confusion.

steve-chavez: >  WDYT? Open for suggestions.

I like it, looks much better now.

> configuration reference page is to get the facts / behaviour exactly right. I didn't touch the tutorials too much on purpose

Agree, the reference doesn't have to include a quickstart, that's better left for the tutorials.

steve-chavez (approved):

PR #494, opened 2022-01-25
Improving docs for setting response headers like content-type

nileshtrivedi:
[I ran into a gotcha](https://github.com/PostgREST/postgrest/issues/2138). Improving docs so that others can avoid this.

This is my first PR in this project, and I haven't tested the generated docs. Feel free to discard if it's not appropriate or not done correctly.

### DIFF
--- a/api.rst
+++ b/api.rst
@@ -2509,7 +2509,7 @@ Notice that the variable should be set to an *array* of single-key objects rathe
 
 .. note::
 
-  PostgREST provided headers such as ``Content-Type``, ``Location``, etc. can be overriden this way. Note that irrespective of overridden ``Content-Type`` response header, the content will still be converted to JSON, unless you also set :ref:`raw_media_types` to something like ``text/html``.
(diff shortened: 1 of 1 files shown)


api.rst
```diff
@@ -2509,7 +2509,7 @@ Notice that the variable should be set to an *array* of single-key objects rathe
 
 .. note::
 
-  PostgREST provided headers such as ``Content-Type``, ``Location``, etc. can be overriden this way.
+  PostgREST provided headers such as ``Content-Type``, ``Location``, etc. can be overriden this way. Note that irrespective of overridden ``Content-Type`` response header, the content will still be converted to JSON, unless you also set :ref:`raw_media_types` to something like ``text/html``.
```
steve-chavez, line 2512: Thanks! The correct link should be:
```suggestion
  PostgREST provided headers such as ``Content-Type``, ``Location``, etc. can be overriden this way. Note that irrespective of overridden ``Content-Type`` response header, the content will still be converted to JSON, unless you also set :ref:`raw-media-types` to something like ``text/html``.
```

PR #495, opened 2022-01-30
ci: switch to GitHub Actions

wolfgangwalther:
Will push this to the other active branches, too (v9.0, v8.0, v7.0.0) - and then try to disable circle CI.

Issue #393, opened 2021-04-05
Translations


wolfgangwalther:
Follow up on #375.

Background: We had a semi-finished translation for chinese in the past. However, this was in a different repo and therefore hard to keep up to date.

The "proper" way of translating a sphinx-based project, would be to set up a PO-file-based translation workflow.

Some resources:
- Sphinx docs: https://www.sphinx-doc.org/en/master/usage/advanced/intl.html
- sphinx-intl: https://github.com/sphinx-doc/sphinx-intl

Then we could connect to a web-based translation platform for contributors to work on, e.g:
- https://poeditor.com/pricing/ (free for open source, we qualify)
- https://www.transifex.com/pricing/ (free for open source, not sure if we qualify)
- others?

The idea would be to set up a basic workflow and maybe translate the index page to a few languages first. This could serve as a localized landing page and also as a proof-of-concept for the workflow.

Down the road this would then enable contributors to work on a full translation, if they want to.

cnyyk:
I'll try to translate to Simplified Chinese.

cnyyk:
Sphinx-intl workflow is ready.

![WX20220110-222636](https://user-images.githubusercontent.com/275233/148782065-9f41b8f5-16f8-4bd8-a999-8e0595a99058.png)

![WX20220110-221643](https://user-images.githubusercontent.com/275233/148782098-aaefd1b7-b118-4ac9-85a4-6aee07717f48.png)

![WX20220110-221706](https://user-images.githubusercontent.com/275233/148782153-5f260907-aeaa-4b15-83dc-061367013c20.png)

cnyyk:
@wolfgangwalther 
There's a problem leaved:
Release notes are separated  html files located in ```releases``` folder.
Should be change to ```rst``` file for translation.

wolfgangwalther:
> Release notes are separated html files located in `releases` folder.
> Should be change to `rst` file for translation.

Those files are `.rst` for me - not `.html`. Not sure what you're looking at exactly?

wolfgangwalther:
> Sphinx-intl workflow is ready.

Thanks for picking this up, this is awesome.

I have not looked at things in detail, yet, but it would be great if we could wrap the tooling that is needed into a few simple nix commands (see https://github.com/PostgREST/postgrest-docs/blob/main/default.nix), that can then be run in `nix-shell`.

Would you like to give this a try?

cnyyk:
[...]

PR #496, opened 2022-01-30
Tooling for translations

wolfgangwalther:
As #393 recently got some traction, I'm adding support for translations to our build tools.

The main features are:
- `postgrest-docs-build` will now update `.po` files' msgids automatically.
- `postgrest-docs-build` and `postgrest-docs-serve` can take an argument to build/serve in a specific language

@cnyyk could you test the `livereload_docs.py` script? It should work without `nix`, too. Just call it like `./livereload_docs.py zh_CN`.

I also moved all the actual docs files to a `docs/` subfolder to separate the tooling from the content a little bit more.

TODO:
- [x] Add CI check to confirm that all changes to `.po` files have been committed. Will do after #495 is done.
- [ ] ~Support `postgrest-docs-spellcheck` for other locales.~ Edit: `aspell` does not support Chinese: http://aspell.net/0.60.7/man-html/Unsupported.html. We might add it later for other languages, once there is a need.

Each translation/locale needs another readthedocs project, which will then be set as a "translation project" in their settings. I created a project for simplified chinese as that's currently worked on. @steve-chavez I added you to the project on readthedocs, too - could you please setup the integration / sync the webhook (I'm missing permissions), so that we can test the build process. I'd like to know how PRs are built once we have a translation. I hope to have builds for both languages at the same time.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,6 +19,7 @@ jobs:
     - uses: cachix/install-nix-action@v25
     - run: nix-env -f default.nix -iA build
     - run: postgrest-docs-build
+    - run: git diff --exit-code HEAD locales || echo "Please commit changes to the locales/ folder after running postgrest-docs-build."
 
   spellcheck:
     name: Run spellcheck
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ diagrams/db.pdf
 misspellings
 unuseddict
 .history
+*.mo
--- a/README.md
+++ b/README.md
@@ -18,3 +18,21 @@ Once in the nix-shell you have the following commands available:
 
 This documentation is structured according to tutorials-howtos-topics-references. For more details on the rationale of this structure, 
 see https://www.divio.com/blog/documentation.
+
+## Translating
+
+To create `.po` files for translation into a new language pass the language code as the first argument to `postgrest-docs-build`.
+
+Example to add German/de:
+
+```
+postgrest-docs-build de
+```
+
+The livereload server also supports a language/locale argument to show the translated docs during translation:
+
+```
+postgrest-docs-serve de
+```
+
+Spellcheck is currently only available for the default language.
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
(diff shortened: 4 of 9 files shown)


steve-chavez:
> could you please setup the integration / sync the webhook (I'm missing permissions), so that we can test the build process

Sure, I've just hit the resync button. I think it worked because the integration page now shows a 200 activity.

wolfgangwalther:
Sweet - it's building both the english and the chinese version for each PR now.

Not quite perfect with the naming, yet:
```
docs/readthedocs.org:postgrest — Read the Docs build succeeded!
docs/readthedocs.org:postgrest-docs — Read the Docs build succeeded!
```

Not sure where we can change that. The second one definitely seems to be built in Chinese, though :)

wolfgangwalther:
> Not sure where we can change that. The second one definitely seems to be built in Chinese, though :)

Turns out I had to recreate the project on readthedocs to change the "project slug". @steve-chavez could you do me a favor and sync the webhook in the newly created project once more?

That should then display it in PRs as `postgrest-zh-cn` instead of `postgrest-docs`.

steve-chavez:
> could you do me a favor and sync the webhook in the newly created project once more?

Done! Got another 200.

wolfgangwalther:
Ok, that works. Should be good to merge, but I'll keep this open until @cnyyk has maybe tested it a bit, too, and has something to put in a PR. No need to carry the empty set of `.po` files in the meantime.

haowuge:
Hello, is there any progress on the Chinese translation project?
 I think there seems to be a pause?

steve-chavez:
Looks like @cnyyk is inactive now. @haowuge Would you like to maintain the chinese docs?

I don't recall the procedure here. Maybe Wolfgang can give you more details.

haowuge:
@steve-chavez Hello steve-chavez , my English is not very good, I tried to translate, but the effect is not satisfactory

yueim:
Here's the old deadline.
image

I'll make an new plan for this work.

wolfgangwalther:
Because of https://github.com/PostgREST/postgrest/issues/2814, I rebased this PR on latest main, fixed the tooling with new nixpkgs. I will merge this without the empty locale files for zh_CN, which only served as an example.

If the tooling needs any improvements, we can work on those later. Just want to get PRs closed here now.

PR #498, opened 2022-01-30
Move *.rst files to docs/ folder

wolfgangwalther:
As mentioned in #496, this gives a bit more structure to the repo. I think it's much easier to find the actual files to work on, instead of having them and all the scripts and stuff in the root folder.


wolfgangwalther:
I'll make the same changes in the other active branches, so that cherry-picking stuff to multiple branches will be easier to do.

PR #501, opened 2022-02-03
Accessing array items and fields of composite types

wolfgangwalther:
Related to https://github.com/PostgREST/postgrest/pull/2145

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -396,7 +396,7 @@ As mentioned, computed columns do not appear in the output by default. However y
 
 .. important::
 
-  Computed columns must be created under the :ref:`exposed schema <db-schemas>` to be used in this way.
+  Computed columns must be created in the :ref:`exposed schema <db-schemas>` or in a schema in the :ref:`extra search path <db-extra-search-path>` to be used in this way.
 
 Unicode support
 ---------------
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -270,7 +270,7 @@ Casting the columns is possible by suffixing them with the double colon ``::`` p
 .. _json_columns:
 
 JSON Columns
-~~~~~~~~~~~~
+------------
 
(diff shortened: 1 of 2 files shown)


docs/api.rst
```diff
@@ -348,10 +348,15 @@ Note that ``->>`` is used to compare ``blood_type`` as ``text``. To compare with
     { "id": 15, "age": 35 }
   ]
 
+
+.. note::
+
+   When using the ``->`` and ``->>`` operators, PostgREST uses a query like ``to_jsonb(<col>)->'field'``. To make filtering and ordering on those nested fields use an index, the index needs to be created on the same expression, including the ``to_jsonb(...)`` call.
```
steve-chavez, line 354: Aha, this is an important detail that I missed. How about:

```suggestion
.. important::

   When using the ``->`` and ``->>`` operators, PostgREST uses a query like ``to_jsonb(<col>)->'field'``. To make filtering and ordering on those nested fields use an index, the index needs to be created on the same expression, including the ``to_jsonb(...)`` call.
```

steve-chavez: Perhaps a snippet can be added to be more explicit? Like:

```sql
CREATE INDEX ON mytable((to_jsonb(data) -> 'identification' ->> 'registration_number'));
```

wolfgangwalther: > Aha, this is an important detail that I missed.

Yeah. The requirement to add an index on a expression was already the case before when using JSON fields. Now I wonder whether existing indexes for JSON fields need to be changed from `col->'field'`` expressions to `to_jsonb(col)->'field'`. I*m pretty sure the answer is yes. Would you consider this a breaking change?

steve-chavez: Yeah, I think it's a breaking change since it invalidates previous indexes, better to be clear about it.

wolfgangwalther: Hm, ok. So we either revert the commit in the postgrest repo to be able to release a minor release next - or we will make the next release a major release. In that case, I would go ahead an merge some of my other breaking changes that are still open.

steve-chavez: Major release sounds good. Ideally we'd have some patch/minor releases after a major one but reverting the commit seems off since we already made a pre-release.

PR #502, opened 2022-02-08
updated the example config switch

electronoob:
replacing postgrest 2 with postgrest -e


wolfgangwalther:
Thanks for the fix!

wolfgangwalther:
I cherry-picked this change back to v9.0 and v8.0, too - as we introduced `--example` in v8.

electronoob:
happy to help @wolfgangwalther

PR #504, opened 2022-02-09
Add admin section on file descriptors

steve-chavez:
Related to https://github.com/PostgREST/postgrest/pull/2158.

Timeouts can be better explained when the `/metrics` endpoint is available, so I've just scoped this PR to file descriptors.

### DIFF
--- a/docs/admin.rst
+++ b/docs/admin.rst
@@ -256,6 +256,7 @@ Schema Reloading
 
 Changing the schema while the server is running can lead to errors due to a stale schema cache. To learn how to refresh the cache see :ref:`schema_reloading`.
 
+
 Daemonizing
 ===========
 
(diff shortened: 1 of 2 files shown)


docs/admin.rst
```diff
@@ -295,6 +296,17 @@ After that, you can enable the service at boot time and start it with:
   ## For reloading the service
   ## systemctl restart postgrest
 
+File Descriptors
+----------------
+
+File descriptors are kernel resources that are used by HTTP connections(among others). File descriptors are limited per process and depending on the Linux distribution, they can have a default limit value of 1024, 4096, etc.
+When under heavy traffic, PostgREST can reach this limit and start showing errors like ``No file descriptors available``. To clear these errors, you can increase the process file descriptors' limit.
```
wolfgangwalther, line 303:
```suggestion
File descriptors are kernel resources that are used by HTTP connections (among others). File descriptors are limited per process. The kernel default limit is 1024, which is increased in some Linux distributions.
When under heavy traffic, PostgREST can reach this limit and start showing ``No file descriptors available`` errors. To clear these errors, you can increase the process' file descriptor limit.
```

PR #505, opened 2022-02-15
Add example for array and composite type columns

laurenceisla:
Adds examples to the information added in #501.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -374,7 +374,7 @@ The arrow operators are also used for array and composite type columns.
 
   .. code-tab:: http
 
-    GET /http://localhost:3000/countries?select=id,location->>lat,location->>long,primary_language:languages->0&location->lat=gte.19 HTTP/1.1
+    GET /countries?select=id,location->>lat,location->>long,primary_language:languages->0&location->lat=gte.19 HTTP/1.1
 
   .. code-tab:: bash Curl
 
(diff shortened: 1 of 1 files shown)


wolfgangwalther:
Cool, I like it :)

wolfgangwalther (approved):

docs/api.rst
```diff
@@ -348,6 +355,41 @@ Note that ``->>`` is used to compare ``blood_type`` as ``text``. To compare with
...
+
+  CREATE TYPE coordinates (
+    lat decimal(8,6),
+    long decimal(9,6)
+  );
+
+  CREATE TABLE countries (
+    id int,
+    location coordinates,
+    languages varchar[]
```
steve-chavez, line 370: [postgres wiki](https://wiki.postgresql.org/wiki/Don't_Do_This#Don.27t_use_varchar.28n.29_by_default) suggests to use `text` instead:

```suggestion
    languages text[]
```

docs/api.rst
```diff
@@ -348,6 +355,41 @@ Note that ``->>`` is used to compare ``blood_type`` as ``text``. To compare with
     { "id": 15, "age": 35 }
   ]
 
+The arrow operators are also used for array and composite type columns.
```
steve-chavez, line 358: How about a new linkable section for these? 

```suggestion
Composite / Array Columns
-------------------------------------

The arrow operators are also used for array and composite type columns.
```

That way we can showcase the feature better in the release notes.

The above `Array / Composite / JSON Columns` would have to be changed to just `JSON columns`.

steve-chavez (approved):

PR #507, opened 2022-02-22
Backward compatible (server versions < 14) code example for getting claims in SQL

netmilk:
Fixing https://github.com/PostgREST/postgrest-docs/issues/469

### DIFF
--- a/docs/auth.rst
+++ b/docs/auth.rst
@@ -93,9 +93,18 @@ Alternately database roles can represent groups instead of (or in addition to) i
 
 SQL code can access claims through GUC variables set by PostgREST per request. For instance to get the email claim, call this function:
 
+For PostgreSQL server version >= 14
+
 .. code:: sql
 
   current_setting('request.jwt.claims', true)::json->>'email';
+  
+
+For PostgreSQL server version < 14
+
+.. code:: sql
+
+  current_setting('request.jwt.claim.email', true);
 
 This allows JWT generation services to include extra information and your database code to react to it. For instance the RLS example could be modified to use this current_setting rather than current_user. The second 'true' argument tells current_setting to return NULL if the setting is missing from the current configuration.
 


netmilk:
Thank you for the merge! In case it fell through the cracks, I described my motivation for the PR here: https://github.com/PostgREST/postgrest-docs/issues/469#issuecomment-1047844574

Also a humble question: What is the lifecycle of the Postgrest documentation? I can still see the pre-merge version on [postgrest.org](https://github.com/PostgREST/postgrest-docs/issues/469#issuecomment-1047844574)

wolfgangwalther:
It's already online at: https://postgrest.org/en/latest/auth.html#web-users-sharing-role

The latest docs are always merged immediately.

You are probably looking at the stable docs for v9. For those to update, we'd need to backpatch the change. You could open a PR with the same change against the `v9.0` branch.

wolfgangwalther:
I just cherry-picked this onto v9.0, so no need anymore to open up a PR.

netmilk:
Thank you million! 

Looking at the postgrest [source code](https://github.com/PostgREST/postgrest/blob/9639cd57510383bd62094430d27d3934725cf7ec/src/PostgREST/Middleware.hs#L71), I realized this behavior might lead to a forward compatibility issue in my SQL code when I upgrade my Postgres to >=14 unless I specify [`db-use-legacy-gucs`](https://postgrest.org/en/stable/configuration.html#db-use-legacy-gucs) in postgres**t** config. Or in case I don't have control over what version of Postgres my code is run against.

If I assume that correctly, here's one more PR: https://github.com/PostgREST/postgrest-docs/pull/510

wolfgangwalther:
You can **not** avoid a breaking change when upgrading from a pre-v9 PostgREST to PostgreSQL v14.

Since pre-v9 PostgREST does not support PostgreSQL 14, your update path must always be:
- Stick with PG13, update to PostgREST 9+. `db-use-legacy-gucs` default to `true` right now, so you don't need to change the code that uses those GUCs.
- Now update your code to use the new GUC-style. Change `db-use-legacy-gucs=false` at the same time.
- Now upgrade PostgreSQL to v14. Because you already use the new style GUCs, this will not break.

If you don't care about downtime, you can do everything together, of course.

PR #508, opened 2022-02-22
Add documentation on the minimal health check

laurenceisla:
Related to:

- https://github.com/PostgREST/postgrest/pull/2109
- https://github.com/PostgREST/postgrest/pull/2092
- https://github.com/PostgREST/postgrest/pull/2107

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -105,6 +105,7 @@ It's not possible to change :ref:`env_variables_config` for a running process an
 
   The following settings will not be reloaded. You will need to restart PostgREST to change those.
 
+    * :ref:`admin-server-port`
     * :ref:`db-uri`
     * :ref:`db-pool`
     * :ref:`db-pool-timeout`
@@ -145,6 +146,7 @@ List of parameters
 ======================== ======= ================= ==========
 Name                     Type    Default           Reloadable
 ======================== ======= ================= ==========
+admin-server-port        Int
 app.settings.*           String                    Y
 db-anon-role             String                    Y
 db-channel               String  pgrst             Y
@@ -173,6 +175,18 @@ server-port              Int     3000
 server-unix-socket       String
 server-unix-socket-mode  String  660
 ======================== ======= ================= ==========
+
+.. _admin-server-port:
+
+admin-server-port
+-----------------
+
+  =============== =======================
+  **Environment** PGRST_ADMIN_SERVER_PORT
+  **In-Database** pgrst.admin_server_port
+  =============== =======================
+
+Specifies the port for the :ref:`health_check` endpoints.
 
 .. _app.settings.*:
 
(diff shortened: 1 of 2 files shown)


docs/configuration.rst
```diff
@@ -174,6 +176,18 @@ server-unix-socket       String
...
 ======================== ======= ================= ==========
 
+.. _admin-server-port:
+
+admin-server-port
+-----------------
+
+  =============== =======================
+  **Environment** PGRST_ADMIN_SERVER_PORT
+  **In-Database** pgrst.admin_server_port
```
steve-chavez, line 186: Shouldn't this be:

```suggestion
  **In-Database** n/a
```

Confirmation pending, not sure.

steve-chavez: `server-host` should likely be the same

laurenceisla: It worked when I tested it. For instance, doing this in the db:

```postgresql
ALTER ROLE authenticator SET pgrst.admin_server_port = "9191";
```

Made the health check service available at port `9191`. I also made sure that the configuration file did not have that setting.

steve-chavez (approved):

docs/configuration.rst
wolfgangwalther: Well.. but we should not document it like this. I consider this a bug, because we just forgot to add it here:

https://github.com/PostgREST/postgrest/blob/dcf7ade5bc70a31fe0e083e1527b74125a94aeda/src/PostgREST/Config.hs#L350-L354

Setting the admin port via DB makes no sense.

wolfgangwalther: > server-host should likely be the same

Seems like I made a mistake there, because all the `server-` variables should certainly **not** have In-Database variants in the docs.

steve-chavez: > Setting the admin port via DB makes no sense.

Hm, it's not reloadable but it works for setting the value at startup, isn't that useful? Same for `server-host`(not reloadable but "initializable").

wolfgangwalther: I don't think those "infrastructure" variables should be set from the database. Everything that is about the database itself and the api that it exposes makes sense to set in the database.

But this is about the environment around it. The database / the schema should have no business which ports are free on the machine and which are not.

Also the difference between initializable and reloadable is tough. It makes thing much easier to reason about when **all** database settings are reloadable.

wolfgangwalther: > The database / the schema should have no business which ports are free on the machine and which are not.

Especially because the database could be on a completely different machine than PostgREST. The server-host, server-port etc. are all about the machine, though - not the database.

steve-chavez: Makes a lot sense. Let's remove the admin port from the in-db config.

laurenceisla: Alright, fixed all the in-db variables here and I'll exclude the setting in the core repo while I'm at it.

Issue #426, opened 2021-08-13
Document special values regarding timestamptz filters


steve-chavez:
Adding the timezone to the value works as discovered on [gitter](https://gitter.im/begriffs/postgrest?at=6116ac618fc359158c6b8e85).

```http
GET /articleStars?createdAt=lt.2020-01-01+00:00:00+Australia/Sydney
```

Using special values such as "now" or "yesterday" also work([ref1](https://github.com/PostgREST/postgrest/issues/1597#issuecomment-701529949), [ref2](https://github.com/PostgREST/postgrest/discussions/1650#discussioncomment-248566)).

steve-chavez:
I think this + PostGIS/hstore(https://github.com/PostgREST/postgrest-docs/issues/273) would fit good on a "Working with types" page.

laurenceisla:
We could also add a case with custom types with a string array field, for instance:

```sql
create type strarr as (a text[]);

create function api.fn_strarr(m strarr) returns strarr as $$
  SELECT m;
$$ LANGUAGE SQL;

select * from api.fn_strarr( '("{a,b,c}")');
```

```bash
curl http://localhost:3000/rpc/fn_strarr \
  -X POST -H "Content-Type: application/json" \
  -d '{ "m": "(\"{a,b,c}\")" }'
```

From this comment https://gitter.im/begriffs/postgrest?at=61a4cee6abdd6644e3aef823

PR #509, opened 2022-02-25
Add how-to for working with PostgreSQL data types

laurenceisla:
Resolves part of #426


docs/how-tos/working-with-postgresql-data-types.rst
```diff
@@ -0,0 +1,96 @@
+Working with PostgreSQL data types
+==================================
+
+You may be wondering if PostgREST allows the use of special literal values, such as ``now`` for timestamps, ``yes`` for booleans or time values including the time zone. The answer is yes to all of them, that is because PostgREST makes use of PostgreSQL text representation to work with data types.
+
+Timestamps
+----------
+
+As mentioned above, you can use the time zone to filter or send data if needed. Let's use this table as an example:
```
steve-chavez, line 9: This connects nicely with the previous paragraph but I think each type section should remain independent, in case reordering happens.
```suggestion
You can use the time zone to filter or send data if needed. Let's use this table as an example:
```

docs/how-tos/working-with-postgresql-data-types.rst
```diff
@@ -0,0 +1,153 @@
+.. _working_with_types:
+
+Working with PostgreSQL data types
+==================================
+
+You may be wondering if PostgREST allows the use of special literal values, such as ``now`` for timestamps, ``yes`` for booleans or time values including the time zone. The answer is yes to all of them, that is because PostgREST makes use of PostgreSQL text representation to work with data types.
```
steve-chavez, line 6: This reads nicely but most users probably don't know about text representation in advance.

How about:

```suggestion
PostgREST makes use of PostgreSQL text representation to work with data types. Thanks to this, you can use special literal values, such as ``now`` for timestamps, ``yes`` for booleans or time values including the time zones. This page describes how you can use literal values for different PostgreSQL types.
```

Paragraph could be refined.

docs/how-tos/working-with-postgresql-data-types.rst
```diff
@@ -0,0 +1,153 @@
...
+.. code-block:: json
+
+  [
+    {
+      "id": 2,
+      "due_date": "2022-02-27T06:00:00-05:00"
+    }
+  ]
+
+But what if, for any reason, you need to cast the ``today`` special value to a timezone different from the server. That would mean using the ``AT TIME ZONE`` construct, which can't be done directly in PostgREST, so your best bet would be to create a view, a :ref:`function <s_procs>` or :ref:`computed columns <computed_cols>`. Let's use the last one for the next example:
```
steve-chavez, line 82: I'd say that's a limitation of the literal value in PostgreSQL, so not directly related to PostgREST. So this would be better phrased as a limitation of the text representation.

steve-chavez: Btw, I've just tested this and it's not true, I can do:

```
select 'today Australia/Sydney'::timestamptz;

      timestamptz
------------------------
 2022-03-02 08:00:00-05
```

It is the `now` literal the one that doesn't work with a time zone.

laurenceisla: > Btw, I've just tested this and it's not true

You're right about the `timestamptz`. I also made a mistake: `due_date` must be of type `timestamp` without time zone.

Thinking about it I'm starting to doubt this use case, due to the fact that the dates are always saved in UTC in the database, there shouldn't be a need to cast `now` to a time zone unless the values inserted in the table are in that same time zone (implying the data was saved incorrectly). Maybe there's a better use case to use as an example for this?

steve-chavez: > implying the data was saved incorrectly

Yeah, I also think is a weird use case this one. How about just leaving this part out then? Then the timestamp section will be short and sweet.

docs/how-tos/working-with-postgresql-data-types.rst
```diff
@@ -0,0 +1,121 @@
+.. _working_with_types:
+
+Working with PostgreSQL data types
+==================================
+
+PostgREST makes use of PostgreSQL text representation to work with data types. Thanks to this, you can use special literal values, such as ``now`` for timestamps, ``yes`` for booleans or time values including the time zones. This page describes how you can use literal values for different PostgreSQL types.
```
steve-chavez, line 6: Btw, TIL about 'yes', 'on' for booleans:

https://www.postgresql.org/docs/current/datatype-boolean.html

That page mentions the "string representation" keyword, perhaps we should use that one instead of "text representation"? The latter could be confusing since there's already a "text" type.

steve-chavez: So like:

```suggestion
PostgREST makes use of PostgreSQL string representations to work with data types. Thanks to this, you can use special values, such as ``now`` for timestamps, ``yes`` for booleans or time values including the time zones. This page describes how you can take advantage of these string representations to perform operations on different PostgreSQL data types.
```

steve-chavez (approved):

PR #510, opened 2022-03-01
SQL using GUC variables forward compatibility note

netmilk:
This is a follow up on the legacy GUC variables doc conversation here: https://github.com/PostgREST/postgrest-docs/pull/507

### DIFF
--- a/docs/auth.rst
+++ b/docs/auth.rst
@@ -100,13 +100,15 @@ For PostgreSQL server version >= 14
   current_setting('request.jwt.claims', true)::json->>'email';
   
 
-For PostgreSQL server version < 14
+For PostgreSQL server version < 14 (can be enforced by `db-use-legacy-gucs <https://postgrest.org/en/stable/configuration.html#db-use-legacy-gucs>`_ config for forward compatibility of your SQL code )
 
 .. code:: sql
 
   current_setting('request.jwt.claim.email', true);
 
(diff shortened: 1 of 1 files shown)


docs/auth.rst
```diff
@@ -100,14 +100,16 @@ For PostgreSQL server version >= 14
   current_setting('request.jwt.claims', true)::json->>'email';
   
 
-For PostgreSQL server version < 14
+For PostgreSQL server version < 14 (can be enforced by `db-use-legacy-gucs <https://postgrest.org/en/stable/configuration.html#db-use-legacy-gucs>`_ config for forward compatibility of your SQL code )
```
steve-chavez, line 103: Hm, according to https://postgrest.org/en/stable/configuration.html#db-use-legacy-gucs:

> pgrst.db_use_legacy_gucs
> On PostgreSQL versions 14 and above, this parameter is ignored.

IIUC, "enforcing" would not be true. The only way to maintain forward compat would be to set db-use-legacy-gucs=false on old versions and use the new GUC notation for old and new pg versions.

wolfgangwalther: Correct. So the comment actually needs to be at the `PostgreSQL server version >= 14` line. The new style **can** be enforced for PG < 14.

steve-chavez:
This looks outdated now. Please reopen if you think it has to be included.

PR #514, opened 2022-03-11
Add information on specifying hostnames for multiple instances

laurenceisla:
Related to https://github.com/PostgREST/postgrest/pull/2182#issuecomment-1063971738

### DIFF
--- a/docs/admin.rst
+++ b/docs/admin.rst
@@ -285,7 +285,7 @@ For instance, to verify if PostgREST is running at ``localhost:3000`` while the 
 
   HTTP/1.1 200 OK
 
- If you are using multiple network interfaces, you need to specify a unique :ref:`hostname <server-host>` in the configuration of each PostgREST instance, this way the admin app can identify the the main app instance correctly.
+If you are using multiple network interfaces, you need to specify a unique :ref:`hostname <server-host>` in the configuration of each PostgREST instance, this way the admin app can identify the the main app instance correctly.
 
 Daemonizing
 ===========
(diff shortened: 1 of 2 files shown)


docs/admin.rst
```diff
@@ -285,6 +285,7 @@ For instance, to verify if PostgREST is running at ``localhost:3000`` while the
 
   HTTP/1.1 200 OK
 
+If you are using multiple network interfaces, you need to specify a unique :ref:`hostname <server-host>` in the configuration of each PostgREST instance, this way the admin app can identify the the main app instance correctly.
```
steve-chavez, line 288:
```suggestion
If you have a machine with multiple network interfaces and multiple PostgREST instances, you need to specify a unique :ref:`hostname <server-host>` in the configuration of each PostgREST instance for the health check to work correctly. Don't use the special values(``!4``, ``*``, etc) in this case because the health check could report a false positive.
```

docs/configuration.rst
```diff
@@ -614,6 +614,8 @@ server-host
   * :code:`*6` - any IPv4 or IPv6 hostname, IPv6 preferred
   * :code:`!6` - any IPv6 hostname
 
+  If you are using multiple network interfaces, you need to specify a unique hostname in the configuration of each PostgREST instance.
```
steve-chavez, line 617: Hm, it's not really true as the instances could work anyway. I think we should leave this part out. The most important thing is to point the multiple network card limitation on the health check.

```suggestion
```

Issue #273, opened 2019-11-07
how-to for INSERTing a PostGIS geometry without a TRIGGER


steve-chavez:
The idea is to use the geometry literal representation:

```sql
select 'SRID=4326;POINT(54.729142 25.268204)'::geometry
-- result: 0101000020E6100000ED116A86545D4B4009A87004A9443940 
```

So, simply:

```sql
create table bla(a geometry);

http POST localhost:3000/bla << JSON
{"a": "SRID=4326;POINT(54.729142 25.268204)"}
JSON
```

No need for a TRIGGER!

(This comes from https://gitter.im/begriffs/postgrest?at=5dc46507a03ae1584f85d77b)

steve-chavez:
Found the idea on https://gis.stackexchange.com/a/68756/153225.

steve-chavez:
Same question on stackoverflow: https://stackoverflow.com/questions/62221504/insert-update-postgis-geometry-column-with-postgrest/62246461#62246461.

Also, wonder if we should have a SO link somewhere, similarly to: https://dhall-lang.org/

steve-chavez:
Turns out that `hstore` also needs something similar:

```sql
create table test.hstore_test(id int, settings hstore);

http POST localhost:3000/hstore_test <<JSON
{"id": 1, "settings": "ORGCTE => DE, ORGNAME => Bundesamt"}
JSON

http GET localhost:3000/hstore_test
[
    {
        "id": 1,
        "settings": {
            "ORGCTE": "DE",
            "ORGNAME": "Bundesamt"
        }
    }
]
```

(Got this thanks to Oleg from https://gnuhost.eu/)

PR #516, opened 2022-03-17
Add how to on PostGIS and hstore data types

laurenceisla:
Closes https://github.com/PostgREST/postgrest-docs/issues/273

### DIFF
--- a/docs/how-tos/working-with-postgresql-data-types.rst
+++ b/docs/how-tos/working-with-postgresql-data-types.rst
@@ -78,3 +78,61 @@ You can use other comparative filters and also `PostgreSQL special date/time inp
       "due_date": "2022-02-27T06:00:00-05:00"
     }
   ]
+
+hstore
+------
+
+You can also work with data types belonging to additional supplied modules such as `hstore <https://www.postgresql.org/docs/current/hstore.html>`_. Let's use the following table:
+
+.. code-block:: postgres
+
+  -- Activate the hstore module in the current database
+  create extension if not exists hstore;
+
+  create table api.countries (
+    id int primary key,
+    name hstore unique
+  );
+
+The ``name`` column will have the name of the country in different formats. You can insert values using the text representation for that data type, for instance:
+
+.. tabs::
+
+  .. code-tab:: http
+
+    POST /countries HTTP/1.1
+    Content-Type: application/json
+
+    [
+      { "id": 1, "name": "common => Egypt, official => \"Arab Republic of Egypt\", native => مصر" },
+      { "id": 2, "name": "common => Germany, official => \"Federal Republic of Germany\", native => Deutschland" }
+    ]
+
+  .. code-tab:: bash Curl
+
+    curl "http://localhost:3000/countries" \
+      -X POST -H "Content-Type: application/json" \
+      -d @- << EOF
+      [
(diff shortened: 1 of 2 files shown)


docs/how-tos/working-with-postgresql-data-types.rst
```diff
@@ -78,3 +78,217 @@ You can use other comparative filters and also `PostgreSQL special date/time inp
...
+
+  -- Activate the hstore module in the current database
+  create extension if not exists hstore;
+
+  create table api.countries (
+    id int primary key,
+    name hstore unique
+  );
+
+The ``name`` column will have the name of the country in different formats. You can insert values using the text representation for that data type, for instance:
```
steve-chavez, line 97:
```suggestion
The ``name`` column will have the name of the country in different formats. You can insert values using the string representation for that data type, for instance:
```

docs/how-tos/working-with-postgresql-data-types.rst
```diff
@@ -78,3 +78,217 @@ You can use other comparative filters and also `PostgreSQL special date/time inp
...
+    curl "http://localhost:3000/countries?select=name->>native&name->>common=like.Egypt"
+
+.. code-block:: json
+
+  [{ "native": "مصر" }]
+
+PostGIS data types
+------------------
+
+You can use the text representation for `PostGIS <https://postgis.net/>`_ data types such as ``geometry`` or ``geography``. As an example, let's create a table using the ``geometry`` type (you need to `install PostGIS <https://postgis.net/install/>`_ first).
```
steve-chavez, line 143:
```suggestion
You can use the string representation for `PostGIS <https://postgis.net/>`_ data types such as ``geometry`` or ``geography``. As an example, let's create a table using the ``geometry`` type (you need to `install PostGIS <https://postgis.net/install/>`_ first).
```

docs/how-tos/working-with-postgresql-data-types.rst
```diff
@@ -78,3 +78,217 @@ You can use other comparative filters and also `PostgreSQL special date/time inp
...
+  -- Activate the postgis module in the current database
+  create extension if not exists postgis;
+
+  create table api.coverage (
+    id int primary key,
+    name text unique,
+    area geometry
+  );
+
+Say you want to add areas in polygon format. The request using text representation would look like:
```
steve-chavez, line 156:
```suggestion
Say you want to add areas in polygon format. The request using string representation would look like:
```

docs/how-tos/working-with-postgresql-data-types.rst
```diff
@@ -78,3 +78,217 @@ You can use other comparative filters and also `PostgreSQL special date/time inp
...
+
+  .. code-tab:: bash Curl
+
+    curl "http://localhost:3000/countries?select=name->>native&name->>common=like.Egypt"
+
+.. code-block:: json
+
+  [{ "native": "مصر" }]
+
+PostGIS data types
```
steve-chavez, line 140: I think just PostGIS for this heading is good, the section and the URL path already implies we're talking about data types `/working-with-postgresql-data-types.html#PostGIS`
```suggestion
PostGIS
```

steve-chavez: Probably geography/geometry would be more precise for this section.. but users will likely search for PostGIS first.. so seems good.

PR #518, opened 2022-03-22
Add JSON, Composite and Ranges types to working with types how-to

### LINKED ISSUE #426: Document special values regarding timestamptz filters

laurenceisla:
Closes #426

It's mentioned in the issue to use Composite Types with array values, but I think this example is enough because it keeps it simple and the user can infer that array values are allowed in the JSON object.


docs/index.rst
```diff
@@ -204,8 +204,6 @@ These are recipes that'll help you address specific use-cases.
 
    how-tos/*
 
-- :doc:`how-tos/casting-type-to-custom-json`
-- :doc:`how-tos/embedding-table-from-another-schema`
```
steve-chavez, line 208: With the removal of these the how-to's are now much more useful! :100:

docs/index.rst
```diff
@@ -204,8 +204,6 @@ These are recipes that'll help you address specific use-cases.
...
 - :doc:`how-tos/providing-images-for-img`
 - `How PostgreSQL triggers work when called with a PostgREST PATCH HTTP request <https://blog.fgribreau.com/2020/11/how-postgresql-triggers-works-when.html>`_
```
steve-chavez, line 208: How about moving this one to the ecosystem page? We can have a `Community How-to's` section there.
```suggestion
```

docs/index.rst
steve-chavez: On the release page, we should document the moving of the `casting-type-to-custom-json` page to the WWTypes page.

docs/index.rst
steve-chavez: Yep, that's fine as well

steve-chavez (approved):

PR #519, opened 2022-03-24
Docs for limited updates/deletions

steve-chavez:
For https://github.com/PostgREST/postgrest/pull/2195

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1479,10 +1479,10 @@ Hints also work alongside ``!inner`` if a top level filtering is needed. From th
 
     curl "http://localhost:3000/orders?select=*,central_addresses!billing_address!inner(*)&central_addresses.code=AB1000"
 
-.. _insert_update:
+.. _insert:
 
-Insertions / Updates
-====================
+Insertions
+==========
 
 All tables and `auto-updatable views <https://www.postgresql.org/docs/current/sql-createview.html#SQL-CREATEVIEW-UPDATABLE-VIEWS>`_ can be modified through the API, subject to permissions of the requester's database role.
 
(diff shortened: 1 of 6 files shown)


wolfgangwalther:
Somehow the headlines / TOC / menu needs a bit more cleanup. "UPSERT" is now below "Specifying column", that seems strange?

steve-chavez:
> Somehow the headlines / TOC / menu needs a bit more cleanup. "UPSERT" is now below "Specifying column", that seems strange?

My bad there, just fixed it.

docs/api.rst
```diff
@@ -1780,6 +1784,45 @@ Deletions also support :code:`Prefer: return=representation` plus :ref:`v_filter
...
+Using ``offset`` to target a different subset of rows is also possible.
+
+.. note::
+
+  This feature obeys the following rules:
+
+   - If the table has a PK(and it's available on the :ref:`schema_cache`), it will use the PK to order and limit the rows. This requires a SELECT privilege on the PK columns.
+   - If no PK is available, it will fallback to using the table's `ctid <https://www.postgresql.org/docs/current/ddl-system-columns.html>`_ for the same purpose. This requires a SELECT privilege on the ``ctid``.
+   - It will work on a view if the PK or `ctid <https://www.postgresql.org/docs/current/ddl-system-columns.html>`_ is included as one of its columns.
+
```
wolfgangwalther, line 1825:
```suggestion
  This will only work, if at least one of the following applies:

   - The table has a PK, which can be used to limit the rows. This requires a SELECT privilege on the PK columns.
   - The target is a table, in which case the `ctid <https://www.postgresql.org/docs/current/ddl-system-columns.html>`_ column can be used for the same purpose. This requires a SELECT privilege on ``ctid``.
   - The target is a view, which includes either the table's PK or `ctid <https://www.postgresql.org/docs/current/ddl-system-columns.html>`_ column as one of its columns.
   
```

I have two questions at this point:
- Is granting column level select privileges on the ctid column possible? Or is this something to have by default?
- The view-thing reads like it could be a challenge to implement right. Will take a careful look in the implementation PR.

steve-chavez: > Is granting column level select privileges on the ctid column possible?

Yes, it's possible but a little weird since it's a hidden column: `GRANT SELECT(ctid) ON TABLE tbl`. That's why I was thinking in the implementation to leave the `ctid` option out but it's just too good that it works on all tables.

Of course a SELECT on all columns would work as well.

steve-chavez: > The view-thing reads like it could be a challenge to implement right

As long as the PK is detected on the view it should be fine, this is probably the most common case.

steve-chavez: This is now updated to match https://github.com/PostgREST/postgrest/pull/2211

steve-chavez:
This is now updated to match https://github.com/PostgREST/postgrest/pull/2211.

The CI linkcheck is failing because of https://scoop.sh, which is a recent failure - if it's still down after some time I'll remove it.

steve-chavez:
> The CI linkcheck is failing because of https://scoop.sh/, which is a recent failure - if it's still down after some time I'll remove it.

Moved forward with that one and just put the Github link instead.

PR #520, opened 2022-03-24
Fix unordered list styles

laurenceisla:
Fix bullet points and separations not showing correctly for unordered lists. Pinned package versions following this PR as reference https://github.com/datalad/datalad/pull/6346


steve-chavez:
Nice, I see the bullets now showing at https://postgrest.org/en/latest/api.html#response-format

PR #524, opened 2022-04-05
feat(api): add match/imatch operators for regular expression support

enote-kane:
refs PostgREST/postgrest#2236

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -69,6 +69,8 @@ lte           :code:`<=`                less than or equal
 neq           :code:`<>` or :code:`!=`  not equal
 like          :code:`LIKE`              LIKE operator (use * in place of %)
 ilike         :code:`ILIKE`             ILIKE operator (use * in place of %)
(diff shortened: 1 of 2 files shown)


docs/api.rst
```diff
@@ -69,6 +69,8 @@ lte           :code:`<=`                less than or equal
 neq           :code:`<>` or :code:`!=`  not equal
 like          :code:`LIKE`              LIKE operator (use * in place of %)
 ilike         :code:`ILIKE`             ILIKE operator (use * in place of %)
+match         :code:`~`                 ~ operator
+imatch        :code:`~*`                ~* operator
```
steve-chavez, line 73: A link to https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-POSIX-REGEXP would be nice here.

enote-kane: Started a small section for pattern matching, with references to the Pattern Matching official docs as well as the index section in module pg_trgm.

steve-chavez (approved):

PR #526, opened 2022-04-10
Add  postgrest-py and remove postgrest-pyclient

J0:
## Remove  outdated links and add postgrest-py
 `datrium/postgrest-pyclient <https://github.com/datrium/postgrest-pyclient>`_ - Python -> Link Invalid


wolfgangwalther:
This is about v7.0 of the docs. We are not building this actively anymore - it's not "public" on readthedocs anymore. That means it's still accessible right now, but not available in the menu etc.

Once we release the next major version (10), the docs for v7.0 will be disabled entirely and will redirect to the stable version.

PR #533, opened 2022-05-06
only use hint for disambiguation

steve-chavez:
Docs for https://github.com/PostgREST/postgrest/pull/2272

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1334,11 +1334,6 @@ For doing resource embedding, PostgREST infers the relationship between two tabl
 However, in cases where there's more than one foreign key between two tables, it's not possible to infer the relationship unambiguously
 by just specifying the tables names.
 
-.. _target_disamb:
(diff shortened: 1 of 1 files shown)


steve-chavez:
https://postgrest--533.org.readthedocs.build/en/533/api.html#embedding-disambiguation

Seeing that now we prefer the hint disambiguation, we go from this:

```http
GET /orders?select=name,billing_address(name) 
```
To this:
```http
GET /orders?select=name,billing_address:addresses!billing_address(name) 
```

Which makes me realize renaming the FK doesn't really make things clearer now, so I think that part should be omitted.

We can mention that the FK name can be used in composite foreign keys cases and just put the autogenerated name.

steve-chavez:
This tried to hide the "target disambiguation" feature - not good since it's still supported.

We should deprecate that form in a new release instead.

PR #536, opened 2022-05-18
Add documentation on embedding views containing joins

### LINKED ISSUE #382: embedding on views: clarify the source columns when using JOINs on the view
From the discussion on https://github.com/PostgREST/postgrest/issues/1230#issuecomment-742400809.

Should be added on https://postgrest.org/en/v7.0.0/api.html#embedding-views

The section could be called **Embedding views with joins**. It should make clear that in case of aliases, the source table comes from the aliased table.

laurenceisla:
Closes #382.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1227,6 +1227,58 @@ It's also possible to embed `Materialized Views <https://www.postgresql.org/docs
 
   If view definitions change you must refresh PostgREST's schema cache for this to work properly. See the section :ref:`schema_reloading`.
 
+Embedding Views Containing Joins
+--------------------------------
+
+The same embedding rules applies to views with joined tables, that is, the column of any table participating in the join with a **foreign key** definition must be present in the view.
+
+For instance, let's create a view called ``film_competition_view`` that will join three tables:
+
+.. code-block:: postgres
+
+  CREATE VIEW film_competition_view AS
+  SELECT
+   films.title as film_title
+   , competitions.name as competition_name
+   , film_id
+  FROM
+   nominations
+   JOIN films ON films.id = nominations.film_id
+   JOIN competitions ON competitions.id = nominations.competition_id;
+
+The view column can embed the table ``films`` because it has the ``film_id`` column that has a foreign key defined.
+
+But what if you want to embed the view with the ``actors`` table? The ``film_id`` column alone will not allow it since it belongs to ``nominations`` and there is no relationship between that table and ``actors`` directly. In this case, you need to add the ``film.id`` column which does have the foreign key to the ``actors`` table.
+
+.. code-block:: postgres
+
+  CREATE VIEW film_competition_view AS
+  SELECT
+   films.title as film_title
+   , competitions.name as competition_name
+   -- To keep the relationship with the films table, this column is renamed
+   , film_id as nominations_film_id
+   -- Adding this column allows the relationships we want
+   , films.id as film_id
+  FROM
+   nominations
+   JOIN films ON films.id = nominations.film_id
+   JOIN competitions ON competitions.id = nominations.competition_id;
+
+Now the view can embed both the ``films`` and ``actors`` tables:
+
+.. tabs::
+
+  .. code-tab:: http
+
+    GET /film_competition_view?select=film_title,films(language),actors(last_name,first_name)&competition_name=ilike.bafta HTTP/1.1
+
+  .. code-tab:: bash Curl
+
+    curl "http://localhost:3000/film_competition_view?select=film_title,films(language),actors(last_name,first_name)&competition_name=ilike.bafta"
+
+As seen above, it is important to verify that the column in the view has the foreign key to the table you want to embed.
+
 .. _embedding_view_chains:
 
 Embedding Chains of Views
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1246,7 +1246,9 @@ For instance, let's create a view called ``film_competition_view`` that will joi
    JOIN films ON films.id = nominations.film_id
    JOIN competitions ON competitions.id = nominations.competition_id;
 
-The view can be embedded with the table ``films`` because it has the ``film_id`` column. But what if you want to embed it with the ``actors`` table, for instance? The ``film_id`` column alone will not allow it since it belongs to ``nominations`` and there is no relationship that can be inferred between that table and ``actors``. In this case, you need to select the ``film.id`` column which will allow any relationship that worked with the ``films`` table (using ``film.id`` as a foreign key) to also work for the view.
+The view can be embedded with the table ``films`` because it has the ``film_id`` column from the ``nominations`` table, which has a relationship with ``films``.
+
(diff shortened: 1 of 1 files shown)


docs/api.rst
```diff
@@ -1227,6 +1227,58 @@ It's also possible to embed `Materialized Views <https://www.postgresql.org/docs
...
+   , competitions.name as competition_name
+   , film_id
+  FROM
+   nominations
+   JOIN films ON films.id = nominations.film_id
+   JOIN competitions ON competitions.id = nominations.competition_id;
+
+The view column can embed the table ``films`` because it has the ``film_id`` column that has a foreign key defined.
+
+But what if you want to embed the view with the ``actors`` table? The ``film_id`` column alone will not allow it since it belongs to ``nominations`` and there is no relationship between that table and ``actors`` directly. In this case, you need to add the ``film.id`` column which does have the foreign key to the ``actors`` table.
```
steve-chavez, line 1251: > ``film.id`` column which does have the foreign key to the ``actors`` table.

The film doesn't have an FK to the `actors` table right? Only an m2m relationship through `roles`(which has the FK)

laurenceisla: True, I clarified the explanation now.

docs/api.rst
```diff
@@ -1227,6 +1227,60 @@ It's also possible to embed `Materialized Views <https://www.postgresql.org/docs
...
+  SELECT
+   films.title as film_title
+   , competitions.name as competition_name
+   , film_id
+  FROM
+   nominations
+   JOIN films ON films.id = nominations.film_id
+   JOIN competitions ON competitions.id = nominations.competition_id;
+
+The view can be embedded with the table ``films`` because it has the ``film_id`` column. But what if you want to embed it with the ``actors`` table, for instance? The ``film_id`` column alone will not allow it since it belongs to ``nominations`` and there is no relationship that can be inferred between that table and ``actors``. In this case, you need to select the ``film.id`` column which will allow any relationship that worked with the ``films`` table (using ``film.id`` as a foreign key) to also work for the view.
```
steve-chavez, line 1249:
```suggestion
The view can be embedded with the table ``films`` because it has the ``film_id`` column.

But what if you want to embed it with the ``actors`` table? The ``film_id`` column alone will not allow it since it belongs to ``nominations``, which doesn't have a relationship with ``actors``. In this case, you need to select the ``film.id`` column, that comes from ``films`` which has a many-to-many relationship with ``actors``.
```

steve-chavez:
```suggestion
The view can be embedded with the table ``films`` because it has the ``film_id`` column, which comes from ``nominations`` and has a relationship with ``films``.

But what if you want to embed it with the ``actors`` table? The ``film_id`` column alone will not allow it since it belongs to ``nominations``, which doesn't have a relationship to``actors``. In this case, you need to select the ``film.id`` column, that comes from ``films`` and has a many-to-many relationship with ``actors``.
```

docs/api.rst
```diff
@@ -1227,6 +1227,58 @@ It's also possible to embed `Materialized Views <https://www.postgresql.org/docs
 
   If view definitions change you must refresh PostgREST's schema cache for this to work properly. See the section :ref:`schema_reloading`.
 
+Embedding Views containing Joins
+--------------------------------
+
+The same embedding rules applies to views with joined tables, that is, the column of any table participating in the join with a **foreign key** definition must be present in the view.
```
steve-chavez, line 1233: On views that contain joins, PostgREST will infer the joined tables' foreign keys based on the source of the columns defined for the view.

docs/api.rst
```diff
@@ -1227,6 +1227,58 @@ It's also possible to embed `Materialized Views <https://www.postgresql.org/docs
...
+
+For instance, let's create a view called ``film_competition_view`` that will join three tables:
+
+.. code-block:: postgres
+
+  CREATE VIEW film_competition_view AS
+  SELECT
+   films.title as film_title
+   , competitions.name as competition_name
+   , film_id
```
steve-chavez, line 1243: Needs alignment
```sql
  SELECT
     films.title as film_title
   , competitions.name as competition_name
   , film_id
```

docs/api.rst
```diff
@@ -1227,6 +1227,58 @@ It's also possible to embed `Materialized Views <https://www.postgresql.org/docs
...
+The view can be embedded with the table ``films`` because it has the ``film_id`` column from the ``nominations`` table, which has a relationship with ``films``.
+
+But what if you want to embed the view with the ``actors`` table? The ``film_id`` column alone will not allow it since it belongs to ``nominations``, which doesn't have a relationship to ``actors``. In this case, you need to select the ``films.id`` column from the ``films`` table, which has a many-to-many relationship with ``actors``.
+
+.. code-block:: postgres
+
+  CREATE VIEW film_competition_view AS
+  SELECT
+   films.title as film_title
+   , competitions.name as competition_name
```
steve-chavez, line 1258:
```sql
  SELECT
     films.title as film_title
   , competitions.name as competition_name
```

docs/api.rst
```diff
@@ -1227,6 +1227,58 @@ It's also possible to embed `Materialized Views <https://www.postgresql.org/docs
...
+   , film_id
+  FROM
+   nominations
+   JOIN films ON films.id = nominations.film_id
+   JOIN competitions ON competitions.id = nominations.competition_id;
```
steve-chavez, line 1247: It's subjective, but I read it better in this way:
```sql
  FROM nominations
  JOIN films ON films.id = nominations.film_id
  JOIN competitions ON competitions.id = nominations.competition_id;
```

docs/api.rst
```diff
@@ -1227,6 +1227,58 @@ It's also possible to embed `Materialized Views <https://www.postgresql.org/docs
...
+   -- To keep the relationship with the films table, this column is renamed
+   , film_id as nominations_film_id
+   -- Adding this column allows the relationships we want
+   , films.id as film_id
+  FROM
+   nominations
+   JOIN films ON films.id = nominations.film_id
+   JOIN competitions ON competitions.id = nominations.competition_id;
+
+Now the view can embed both the ``films`` and ``actors`` tables:
```
steve-chavez, line 1268: Now the view can be embedded with both the ``films`` and ``actors`` tables:

docs/api.rst
```diff
@@ -1227,6 +1227,58 @@ It's also possible to embed `Materialized Views <https://www.postgresql.org/docs
...
+   , competitions.name as competition_name
+   , film_id
+  FROM
+   nominations
+   JOIN films ON films.id = nominations.film_id
+   JOIN competitions ON competitions.id = nominations.competition_id;
+
+The view can be embedded with the table ``films`` because it has the ``film_id`` column from the ``nominations`` table, which has a relationship with ``films``.
+
+But what if you want to embed the view with the ``actors`` table? The ``film_id`` column alone will not allow it since it belongs to ``nominations``, which doesn't have a relationship to ``actors``. In this case, you need to select the ``films.id`` column from the ``films`` table, which has a many-to-many relationship with ``actors``.
```
steve-chavez, line 1251: But what if you want to embed the view with the ``actors`` table? The ``film_id`` column alone will not allow it since it belongs to ``nominations``, which doesn't have a relationship to ``actors``. In this case, you need to select the ``id`` column from the ``films`` table, which has a many-to-many relationship with ``actors``.

PR #537, opened 2022-05-19
Fix main explanation on views with joins

laurenceisla:
Related to https://github.com/PostgREST/postgrest-docs/pull/536#discussion_r876534095

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1230,7 +1230,7 @@ It's also possible to embed `Materialized Views <https://www.postgresql.org/docs
 Embedding Views containing Joins
 --------------------------------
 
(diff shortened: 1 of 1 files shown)


docs/api.rst
```diff
@@ -1230,54 +1230,52 @@ It's also possible to embed `Materialized Views <https://www.postgresql.org/docs
 Embedding Views containing Joins
 --------------------------------
 
-The same embedding rules applies to views with joined tables, that is, the column of any table participating in the join with a **foreign key** definition must be present in the view.
+On views that contain joins, PostgREST will infer the joined tables' foreign keys based on the source of the columns defined for the view.
```
steve-chavez, line 1233:
```suggestion
On views that contain joins, PostgREST will infer the joined tables' foreign keys based on the sources of the view columns.
```

PR #538, opened 2022-05-20
Add how to work with array columns

laurenceisla:
Adds examples for Array types in the "How-to work with PostgreSQL types" page.

### DIFF
--- a/docs/how-tos/working-with-postgresql-data-types.rst
+++ b/docs/how-tos/working-with-postgresql-data-types.rst
@@ -204,7 +204,7 @@ To query the data you can use the arrow operators. See :ref:`composite_array_col
 Multidimensional Arrays
 ~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


docs/how-tos/working-with-postgresql-data-types.rst
```diff
@@ -127,6 +127,135 @@ Now, you can insert a new product using a JSON object for the ``extra_info`` col
...
+        "performance_times": ["12:40", "15:00", "20:00"]
+      }
+    EOF
+
+To query the data you can use the arrow operators. See :ref:`composite_array_columns`.
+
+Multidimensional Arrays
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Handling multidimensional arrays is no different than handling simple ones: both the string representation and the JSON array format are allowed. For example, let's add a new column to the table:
```
steve-chavez, line 207:
```suggestion
Handling multidimensional arrays is no different than handling one-dimensional ones: both the string representation and the JSON array format are allowed. For example, let's add a new column to the table:
```

Issue #327, opened 2020-05-28
Clarify Providing images for <img> how-to


steve-chavez:
Clarify http://postgrest.org/en/v7.0.0/how-tos/providing-images-for-img.html.

Some feedback on https://gitter.im/begriffs/postgrest?at=5eccf5282c49c45f5aa197b4. Not clear that nginx sets the correct headers.

Also maybe add a note with a link to https://wiki.postgresql.org/wiki/BinaryFilesInDB. And mention CDN as an alternative for scaling, see https://github.com/PostgREST/postgrest/issues/1077.

PR #542, opened 2022-05-25
Add bytea to working with types and improve img example

### LINKED ISSUE #539: Use the raw-media-types config in the images how-to
It was suggested here https://gitter.im/begriffs/postgrest?at=6285fe78c61b987d33b763fe that the `raw-media-types` could be used to replace the minimal example on how to serve images instead of using Nginx rewrite rules.

Perhaps see which image types are common for popular browsers and add that to `raw-media-types`, for instance.

May be related to #327.

laurenceisla:
Closes #539 and adds info mentioned in https://github.com/PostgREST/postgrest-docs/issues/539#issuecomment-1136213676

### DIFF
--- a/docs/how-tos/providing-images-for-img.rst
+++ b/docs/how-tos/providing-images-for-img.rst
@@ -29,7 +29,7 @@ That's because browsers do not send the required :code:`Accept: application/octe
 
 Luckily we can specify the accepted media types in the :ref:`raw-media-types` configuration variable.
 In this case, the :code:`Accept: image/webp` header is sent by many web browsers by default, so let's add it to the configuration variable, like this: :code:`raw-media-types="image/webp"`.
-Now, the image will be displayed in the html page:
+Now, the image will be displayed in the HTML page:
 
 .. code-block:: html
 
@@ -85,7 +85,7 @@ For production, you probably want to configure additional caches, e.g. on the :r
      end
    $$ language plpgsql;
 
-With this, we can obtain the cat image from :code:`/rpc/file?id=42`. Thus, the resulting html will be:
+With this, we can obtain the cat image from :code:`/rpc/file?id=42`. Thus, the resulting HTML will be:
 
 .. code-block:: html
 
--- a/docs/how-tos/providing-images-for-img.rst
+++ b/docs/how-tos/providing-images-for-img.rst
@@ -10,9 +10,9 @@ In this how-to, you will learn how to create an endpoint for providing images to
 We will start with a minimal example that highlights the general concept.
 Afterwards we present a more detailed solution that fixes a few shortcomings of the first approach.
 
-.. note::
+.. warning::
 
-   Be careful when saving binaries in the database, you may want to use a separate storage service to handle files. See `the PostgreSQL wiki <https://wiki.postgresql.org/wiki/BinaryFilesInDB>`_ for reference.
+   Be careful when saving binaries in the database, having a separate storage service for these is preferable in most cases. See `Storing Binary files in the Database <https://wiki.postgresql.org/wiki/BinaryFilesInDB>`_.
 
 Minimal Example
 ---------------
--- a/docs/how-tos/providing-images-for-img.rst
+++ b/docs/how-tos/providing-images-for-img.rst
@@ -12,7 +12,7 @@ Afterwards we present a more detailed solution that fixes a few shortcomings of 
 
 .. note::
 
-   There are some caveats when saving images in the database as `mentioned in the PostgreSQL wiki <https://wiki.postgresql.org/wiki/BinaryFilesInDB>`_. Also, consider using a Content Delivery Network (CDN) for scalability purposes, see `this discussion <https://github.com/PostgREST/postgrest/issues/1077>`_ for reference.
+   Be careful when saving binaries in the database, you may want to use a separate storage service to handle files. See `the PostgreSQL wiki <https://wiki.postgresql.org/wiki/BinaryFilesInDB>`_ for reference.
 
 Minimal Example
 ---------------
--- a/docs/how-tos/providing-images-for-img.rst
+++ b/docs/how-tos/providing-images-for-img.rst
@@ -10,6 +10,10 @@ In this how-to, you will learn how to create an endpoint for providing images to
 We will start with a minimal example that highlights the general concept.
 Afterwards we present a more detailed solution that fixes a few shortcomings of the first approach.
 
+.. note::
+
+   There are some caveats when saving images in the database as `mentioned in the PostgreSQL wiki <https://wiki.postgresql.org/wiki/BinaryFilesInDB>`_. Also, consider using a Content Delivery Network (CDN) for scalability purposes, see `this discussion <https://github.com/PostgREST/postgrest/issues/1077>`_ for reference.
+
 Minimal Example
 ---------------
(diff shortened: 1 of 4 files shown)


docs/how-tos/working-with-postgresql-data-types.rst
```diff
@@ -5,6 +5,10 @@ Working with PostgreSQL data types
 
 PostgREST makes use of PostgreSQL string representations to work with data types. Thanks to this, you can use special values, such as ``now`` for timestamps, ``yes`` for booleans or time values including the time zones. This page describes how you can take advantage of these string representations to perform operations on different PostgreSQL data types.
 
+.. contents::
+  :local:
+  :depth: 1
+
```
steve-chavez, line 11: Nice! I thought the TOC needed to be manual.

docs/how-tos/working-with-postgresql-data-types.rst
```diff
@@ -461,6 +465,80 @@ Finally, do the request :ref:`casting the range column <casting_columns>`:
...
+   create table files (
+     id serial primary key,
+     file bytea
+   );
+
+   create function upload_binary(bytea) returns void as $$
+     insert into files (file) values ($1);
+   $$ language sql;
+
+.. TODO: GRANT FOR SERIAL
```
steve-chavez, line 484: You can use the `generated always as identity` instead of `serial`.

laurenceisla: Didn't know that it's even suggested not to use `serial`: https://wiki.postgresql.org/wiki/Don%27t_Do_This#Don.27t_use_serial

docs/how-tos/providing-images-for-img.rst
```diff
@@ -36,64 +25,40 @@ First, we need a public table for storing the files.
...
-       return 404;
-       # request binary output
-       proxy_set_header Accept application/octet-stream;
-       # usual proxy setup
-       proxy_hide_header Content-Location;
-       add_header Content-Location /api/$upstream_http_content_location;
-       proxy_set_header  Connection "";
-       proxy_http_version 1.1;
-       proxy_pass http://localhost:3000/;
-     }
```
laurenceisla, line 64: It looks like the examples do not need the reverse proxy anymore. Maybe add another section below, mentioning that `raw-media-types` won't be needed if the :ref:`Accept: application/octet-stream` header is added in Nginx?

steve-chavez: Oh, so we don't need Nginx at all now right? Looks much better this way.

> that raw-media-types won't be needed if the Accept: application/octet-stream header is added in Nginx?

I think, it's fine to leave Nginx out for this use case.

docs/how-tos/providing-images-for-img.rst
```diff
@@ -5,25 +5,14 @@ Providing images for ``<img>``
...
 
-In this how-to, you will learn how to create an endpoint for providing images to HTML :code:`<img>` tags without client side JavaScript.
-The resulting HTML might look like this:
-
-.. code-block:: html
-
-   <img src="http://host/files/42/cats.jpeg" alt="Cute Kittens"/>
-
-In fact, the presented technique is suitable for providing not only images, but arbitrary files.
+In this how-to, you will learn how to create an endpoint for providing images to HTML :code:`<img>` tags without client side JavaScript. In fact, the presented technique is suitable for providing not only images, but arbitrary files.
```
steve-chavez, line 8: I'd like to add a warning for this how-to, what I mentioned in https://github.com/PostgREST/postgrest-docs/issues/327#issue-626719955.

> Also maybe add a note with a link to https://wiki.postgresql.org/wiki/BinaryFilesInDB
> And mention CDN as an alternative for scaling, see the discussion on https://github.com/PostgREST/postgrest/issues/1077.

Maybe the CDN part can be left out, it's not only about scaling in throughput but pg WALs(backups, replicas) also get bloated with big files.

steve-chavez: Just linking to the discussion would be good though

docs/how-tos/providing-images-for-img.rst
```diff
@@ -5,25 +5,18 @@ Providing images for ``<img>``
...
-   <img src="http://host/files/42/cats.jpeg" alt="Cute Kittens"/>
+We will start with a minimal example that highlights the general concept.
+Afterwards we present a more detailed solution that fixes a few shortcomings of the first approach.
 
-In fact, the presented technique is suitable for providing not only images, but arbitrary files.
+.. note::
 
-We will start with a minimal example that highlights the general concept.
-Afterwards we present are more detailed solution that fixes a few shortcomings of the first approach.
+   There are some caveats when saving images in the database as `mentioned in the PostgreSQL wiki <https://wiki.postgresql.org/wiki/BinaryFilesInDB>`_. Also, consider using a Content Delivery Network (CDN) for scalability purposes, see `this discussion <https://github.com/PostgREST/postgrest/issues/1077>`_ for reference.
```
steve-chavez, line 15: Can't think of good words now, but this needs to be more than "some caveats", almost like "think twice before doing this". Also, the CDN thing should be left out(I don't know how good it is to recommend it).

steve-chavez (approved):

docs/how-tos/working-with-postgresql-data-types.rst
```diff
@@ -461,6 +465,80 @@ Finally, do the request :ref:`casting the range column <casting_columns>`:
 
       create cast (mytsrange as json) with function mytsrange_to_json(mytsrange) as assignment;
 
+Binary Data Types
+-----------------
```
steve-chavez, line 469: How about just bytea?
```suggestion
Bytea
--------
```

docs/how-tos/working-with-postgresql-data-types.rst
```diff
@@ -461,6 +465,80 @@ Finally, do the request :ref:`casting the range column <casting_columns>`:
...
+  .. code-tab:: bash Curl
+
+    curl "http://localhost:3000/files?select=file&id=eq.1" \
+      -H "Accept: image/png"
+
+See :ref:`providing_img` for a step-by-step example on how to handle images in HTML.
+
+.. note::
+
+   Be careful when saving binaries in the database, you may want to use a separate storage service to handle files. See `the PostgreSQL wiki <https://wiki.postgresql.org/wiki/BinaryFilesInDB>`_ for reference.
```
steve-chavez, line 540:
```suggestion
   Be careful when saving binaries in the database, having a separate storage service for these is preferable in most cases. See `the PostgreSQL wiki <https://wiki.postgresql.org/wiki/BinaryFilesInDB>`_ for reference.
```

docs/how-tos/working-with-postgresql-data-types.rst
```diff
@@ -461,6 +465,80 @@ Finally, do the request :ref:`casting the range column <casting_columns>`:
...
+    Accept: image/png
+
+  .. code-tab:: bash Curl
+
+    curl "http://localhost:3000/files?select=file&id=eq.1" \
+      -H "Accept: image/png"
+
+See :ref:`providing_img` for a step-by-step example on how to handle images in HTML.
+
+.. note::
```
steve-chavez, line 538:
```suggestion
.. warning::
```

PR #560, opened 2022-07-18
Reword and clarify embedding on views

laurenceisla:
- Clarify that the tables present in FROM and JOIN must not be sub queries. 
- Rephrase Embedding on Views Using Joins from a how-to wording to be more descriptive.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1204,7 +1204,7 @@ Since it contains the ``films_id`` foreign key, it is possible to embed ``box_of
 Embedding Views
 ---------------
 
-Embedding a view is possible if its columns have **foreign keys** defined in their source tables. Source tables are the ones referenced in the ``FROM`` clause and the multiple ``JOIN`` clauses.
+PostgREST will infer the relationships of a view based on its source tables. Source tables are the ones referenced in the ``FROM`` and ``JOIN`` clauses of the view definition. The foreign keys of the relationships must be present in the top ``SELECT`` clause of the view for this to work.
 
 For instance, the following view has ``nominations``, ``films`` and ``competitions`` as source tables:
 
@@ -1221,7 +1221,7 @@ For instance, the following view has ``nominations``, ``films`` and ``competitio
     JOIN films ON films.id = nominations.film_id
     JOIN competitions ON competitions.id = nominations.competition_id;
 
-Since this view contains ``nominations.film_id`` — with a **foreign key** defined in its source table — then we can embed the ``films`` table. Similarly, because the view contains ``films.id``, then we can also embed the ``roles`` and the ``actors`` tables (the last one in a many-to-many relationship):
+Since this view contains ``nominations.film_id``, which has a **foreign key** relationship to ``films``, then we can embed the ``films`` table. Similarly, because the view contains ``films.id``, then we can also embed the ``roles`` and the ``actors`` tables (the last one in a many-to-many relationship):
 
 .. tabs::
 


docs/api.rst
```diff
@@ -1204,9 +1204,9 @@ Since it contains the ``films_id`` foreign key, it is possible to embed ``box_of
 Embedding Views
 ---------------
 
-Embedding a view is possible if the view contains columns that have **foreign keys** defined in their source tables.
+Embedding a view is possible if its columns have **foreign keys** defined in their source table, which, in turn, must be clearly present in the ``FROM`` clause of the view definition (without using sub queries).
```
steve-chavez, line 1207: I've been thinking of a further simplification. How about if we refer to source tables as the ones in the FROM/JOIN targets? So in `FROM (SELECT * FROM other_table)`, `other_table` wouldn't be a source table for us.

steve-chavez: So it'd be:
```suggestion
Embedding a view is possible if its columns have **foreign keys** defined in their source table. Source tables are the tables referenced in the ``FROM`` and ``JOIN`` clauses.
```

docs/api.rst
```diff
@@ -1204,31 +1204,34 @@ Since it contains the ``films_id`` foreign key, it is possible to embed ``box_of
 Embedding Views
 ---------------
 
-Embedding a view is possible if the view contains columns that have **foreign keys** defined in their source tables.
+Embedding a view is possible if its columns have **foreign keys** defined in their source table. Source tables are the ones referenced in the ``FROM`` clause and the multiple ``JOIN`` clauses.
 
-As an example, let's create a view called ``nominations_view`` based on the *nominations* table.
+For instance, the following view has ``nominations``, ``films``and ``competitions`` as source tables:
```
steve-chavez, line 1209:
```suggestion
For instance, the following view has ``nominations``, ``films`` and ``competitions`` as source tables:
```

docs/api.rst
```diff
@@ -1204,31 +1204,34 @@ Since it contains the ``films_id`` foreign key, it is possible to embed ``box_of
 Embedding Views
 ---------------
 
-Embedding a view is possible if the view contains columns that have **foreign keys** defined in their source tables.
+Embedding a view is possible if its columns have **foreign keys** defined in their source table. Source tables are the ones referenced in the ``FROM`` clause and the multiple ``JOIN`` clauses.
```
steve-chavez, line 1207:
```suggestion
Embedding a view is possible if its columns have **foreign keys** defined in their source tables. Source tables are the ones referenced in the ``FROM`` clause and the multiple ``JOIN`` clauses.
```

docs/api.rst
```diff
@@ -1204,31 +1204,34 @@ Since it contains the ``films_id`` foreign key, it is possible to embed ``box_of
 Embedding Views
 ---------------
 
-Embedding a view is possible if the view contains columns that have **foreign keys** defined in their source tables.
+Embedding a view is possible if its columns have **foreign keys** defined in their source tables. Source tables are the ones referenced in the ``FROM`` clause and the multiple ``JOIN`` clauses.
```
steve-chavez, line 1207:
```suggestion
PostgREST will infer the relationships of a view based on its source tables. Source tables are the ones referenced in the FROM and JOIN clauses of the view definition. The foreign keys of the relationships must be present in the top SELECT clause of the view for this to work.
```

docs/api.rst
```diff
@@ -1204,31 +1204,34 @@ Since it contains the ``films_id`` foreign key, it is possible to embed ``box_of
...
+     , competitions.name as competition_name
+     , nominations.rank
+     , nominations.film_id as nominations_film_id
+     , films.id as film_id
+    FROM nominations
+    JOIN films ON films.id = nominations.film_id
+    JOIN competitions ON competitions.id = nominations.competition_id;
 
-Since it contains ``competition_id`` and ``film_id`` — and each one has a **foreign key** defined in its source table — we can embed *competitions* and *films*:
+Since this view contains ``nominations.film_id`` — with a **foreign key** defined in its source table — then we can embed the ``films`` table. Similarly, because the view contains ``films.id``, then we can also embed the ``roles`` and the ``actors`` tables (the last one in a many-to-many relationship):
```
steve-chavez, line 1224:
```suggestion
Since this view contains ``nominations.film_id``, which has a **foreign key** relationship to ``films``, then we can embed the ``films`` table. Similarly, because the view contains ``films.id``, then we can also embed the ``roles`` and the ``actors`` tables (the last one in a many-to-many relationship):
```

steve-chavez (approved):

PR #564, opened 2022-08-25
For v10

steve-chavez:
- [x] limited udate/delete on release page

~- [ ] hint disambiguation needs to be corrected~

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1531,14 +1531,14 @@ For doing resource embedding, PostgREST infers the relationship between two tabl
 However, in cases where there's more than one foreign key between two tables, it's not possible to infer the relationship unambiguously
 by just specifying the tables names.
 
+For example, suppose you have the following ``orders`` and ``addresses`` tables.
+
+.. image:: _static/orders.png
+
 .. _target_disamb:
 
 Target Disambiguation
 ~~~~~~~~~~~~~~~~~~~~~
-
-For example, suppose you have the following ``orders`` and ``addresses`` tables:
-
-.. image:: _static/orders.png
 
 And you try to embed ``orders`` with ``addresses`` (this is the **target**):
 
@@ -1683,7 +1683,7 @@ Hints also work alongside ``!inner`` if a top level filtering is needed. From th
 
 .. note::
 
-  If the relationship is so complex that hint disambiguation does not solve it, then using :ref:`computed_relationships` is the best alternative.
+  If the relationship is so complex that hint disambiguation does not solve it, you can use :ref:`computed_relationships`.
 
 .. _insert:
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -875,52 +875,15 @@ returned together. For example, consider a database of films and their awards:
 
 .. important::
 
-  * PostgREST needs `FOREIGN KEY constraints <https://www.postgresql.org/docs/current/tutorial-fk.html>`_ to be able to do Resource Embedding.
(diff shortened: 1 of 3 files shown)


docs/api.rst
```diff
@@ -2956,8 +3034,8 @@ The same execution can be returned in ``json`` format by using the ``Accept: app
     }
   ]
 
-You can also get the result plan of the different media types that PostgREST supports by adding them to the header using ``for``. For instance, to obtain the plan for a :ref:`text/xml <scalar_return_formats>` media type in json format, you need to add the ``Accept: application/vnd.pgrst.plan; for=text/xml`` header.
+By default the plan it's assumed to generate the JSON representation of a resource(``application/json``) but you can obtain the plan for the :ref:`different representations that PostgREST supports <res_format>` by adding them to the ``for`` parameter. For instance, to obtain the plan for a ``text/xml``, you would use ``Accept: application/vnd.pgrst.plan; for="text/xml``.
```
laurenceisla, line 3037:
```suggestion
By default the plan is assumed to generate the JSON representation of a resource(``application/json``), but you can obtain the plan for the :ref:`different representations that PostgREST supports <res_format>` by adding them to the ``for`` parameter. For instance, to obtain the plan for a ``text/xml``, you would use ``Accept: application/vnd.pgrst.plan; for="text/xml``.
```

docs/releases/v10.0.0.rst
```diff
@@ -5,79 +5,65 @@ PostgREST 10.0.0
...
   - If :ref:`db-uri` is not set, PostgREST will use the `libpq environment variables <https://www.postgresql.org/docs/current/libpq-envars.html>`_ for the database connection.
   - If :ref:`db-schemas` is not set, it will use the database ``public`` schema.
   - If :ref:`db-anon-role` is not set, it will not allow anonymous requests.
 
+Error messages
+~~~~~~~~~~~~~~
+
+- To increase consistency, all the errors messages are now normalized. The ``hint``, ``details``, ``code`` and ``message`` fields will always be present in the body, each one defaulting to a ``null`` value. In the same way, the :ref:`errors that were raised <raise_error>` with ``SQLSTATE`` now include the ``message`` and ``code`` in the body.
+
+- To further clarify the source of an error, we now add a ``PGRST`` prefix to the error code of all the errors that are PostgREST-specific and don't come from the database. These errors have unique codes that identifies them and are documented in the :ref:`pgrst_errors` section.
```
laurenceisla, line 65:
```suggestion
- To further clarify the source of an error, we now add a ``PGRST`` prefix to the error code of all the errors that are PostgREST-specific and don't come from the database. These errors have unique codes that identify them and are documented in the :ref:`pgrst_errors` section.
```

docs/api.rst
```diff
@@ -1451,15 +1531,15 @@ For doing resource embedding, PostgREST infers the relationship between two tabl
...
 .. _target_disamb:
 
 Target Disambiguation
 ~~~~~~~~~~~~~~~~~~~~~
 
-For example, suppose you have the following ``orders`` and ``addresses`` tables:
-
-.. image:: _static/orders.png
-
 And you try to embed ``orders`` with ``addresses`` (this is the **target**):
```
laurenceisla, line 1543:
```suggestion
Let's try to embed ``orders`` with ``addresses`` (this is the **target**):
```

Or start with something different than "And", since it's not a continuation of the previous sentence anymore.

steve-chavez: Removed the previous change instead

PR #571, opened 2022-09-21
Add information on preflight requests

laurenceisla:
Specify that preflight requests are supported, returning the max-age required for caching.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -2705,7 +2705,9 @@ CORS
 
 PostgREST sets highly permissive cross origin resource sharing, that is why it accepts Ajax requests from any domain.
 
-It also handles `preflight requests <https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request>`_ done by the browser, which can be cached using the returned ``Access-Control-Max-Age`` header. This is useful to reduce the latency of the subsequent requests by approximately 100 ms. A ``POST`` preflight request would look like this:
+It also handles `preflight requests <https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request>`_ done by the browser, which can be cached using the returned ``Access-Control-Max-Age`` header. This is useful to reduce the latency of the subsequent requests.
+
+A ``POST`` preflight request would look like this:
 
 .. tabs::
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -2705,7 +2705,7 @@ CORS
 
 PostgREST sets highly permissive cross origin resource sharing, that is why it accepts Ajax requests from any domain.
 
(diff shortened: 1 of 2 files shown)


docs/api.rst
```diff
@@ -2705,6 +2705,34 @@ CORS
 
 PostgREST sets highly permissive cross origin resource sharing, that is why it accepts Ajax requests from any domain.
 
+It also handles `preflight requests <https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request>`_ done by the browser, which can be cached using the returned ``Access-Control-Max-Age`` header. This is useful to reduce the latency of the subsequent requests by approximately 100 ms. A ``POST`` preflight request would look like this:
```
steve-chavez, line 2708: 100ms - it might be way off depending on the location. So better omit it 

```suggestion
It also handles `preflight requests <https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request>`_ done by the browser, which can be cached using the returned ``Access-Control-Max-Age`` header. This is useful to reduce the latency of the subsequent requests.

A ``POST`` preflight request would look like this:
```

docs/api.rst
```diff
@@ -2705,6 +2705,34 @@ CORS
...
+       -H "Access-Control-Request-Headers: Content-Type"
+
+.. code-block:: http
+
+  HTTP/1.1 200 OK
+  Access-Control-Allow-Origin: http://example.com
+  Access-Control-Allow-Credentials: true
+  Access-Control-Allow-Methods: GET, POST, PATCH, PUT, DELETE, OPTIONS, HEAD
+  Access-Control-Allow-Headers: Authorization, Content-Type, Accept, Accept-Language, Content-Language
+  Access-Control-Max-Age: 86400
```
steve-chavez, line 2734: This should be mentioned in the paragraph above. 86400 seconds = 24 hours.

docs/api.rst
```diff
@@ -2705,6 +2705,36 @@ CORS
 
 PostgREST sets highly permissive cross origin resource sharing, that is why it accepts Ajax requests from any domain.
 
+It also handles `preflight requests <https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request>`_ done by the browser, which can be cached using the returned ``Access-Control-Max-Age: 86400`` header (86400 seconds = 24 hours). This is useful to reduce the latency of the subsequent requests.
```
steve-chavez, line 2708:
```suggestion
It also handles `preflight requests <https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request>`_ done by the browser, which are cached using the returned ``Access-Control-Max-Age: 86400`` header (86400 seconds = 24 hours). This is useful to reduce the latency of the subsequent requests.
```

steve-chavez (approved):

PR #586, opened 2022-12-29
adding link to ops unikernel package

eyberg:


steve-chavez:
To correct the CI error, you can add `Nanos` and `unikernel` here: https://github.com/PostgREST/postgrest-docs/blob/main/postgrest.dict

PR #588, opened 2023-01-29
Add db-plan-enabled recommendation

steve-chavez:
> Perhaps we can recommend an additional measure: a db-pre-request that only enables the plan media type from certain IPs, otherwise rejects it.

Discussed on https://github.com/PostgREST/postgrest/pull/2368#issuecomment-1189816556

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -296,8 +296,8 @@ db-plan-enabled
 
   When this is set to :code:`true`, the execution plan of a request can be retrieved by using the :code:`Accept: application/vnd.pgrst.plan` header. See :ref:`explain_plan`.
 
-  It's recommended to only use this on testing environments and not in production as it reveals internal database details.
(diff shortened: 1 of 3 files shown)


docs/configuration.rst
```diff
@@ -296,6 +296,30 @@ db-plan-enabled
 
   When this is set to :code:`true`, the execution plan of a request can be retrieved by using the :code:`Accept: application/vnd.pgrst.plan` header. See :ref:`explain_plan`.
 
+  It's recommended to only use this on testing environments and not in production as it reveals internal database details.
+  However if you choose to use it in production you can use a :ref:`db-pre-request` to filter the requests that can use this feature.
```
laurenceisla, line 300:
```suggestion
  It's recommended to use this in testing environments only since it reveals internal database details.
  However, if you choose to use it in production you can implement a :ref:`db-pre-request` to filter the requests that can use this feature.
```

I suggested "implement" but any other word can be to replace "use", because it's repeated a lot.

laurenceisla (approved):

PR #604, opened 2023-04-21
Add caveats when working with computed relationships

### LINKED ISSUE #570: Incorrect `ROWS 1` estimation gives unexpected results for computed relationships
Mentioned here: https://github.com/PostgREST/postgrest/issues/2475#issue-1372339547

It should be documented as a warning that when using a computed relationship and the estimation is mistakenly set to `ROWS 1`, then PostgREST will return an unnested left join. For instance, from the above issue, this is an example that is returned:

```http
GET /premieres?select=*,computed_films_m2m_marked_as_m2o(*)

-- expect usage, unexpected results (unnested left join)
[{"id":"P1","film_id":"F1","computed_films_m2m_marked_as_m2o":{"id":"F1"}}, 
 {"id":"P1","film_id":"F1","computed_films_m2m_marked_as_m2o":{"id":"F2"}}, 
 {"id":"P1","film_id":"F1","computed_films_m2m_marked_as_m2o":{"id":"F4"}}, 
 {"id":"P2","film_id":"F1","computed_films_m2m_marked_as_m2o":{"id":"F1"}}, 
 {"id":"P2","film_id":"F1","computed_films_m2m_marked_as_m2o":{"id":"F2"}}, 
 {"id":"P2","film_id":"F1","computed_films_m2m_marked_as_m2o":{"id":"F4"}}, 
 {"id":"P3","film_id":"F2","computed_films_m2m_marked_as_m2o":{"id":"F1"}}, 
 {"id":"P3","film_id":"F2","computed_films_m2m_marked_as_m2o":{"id":"F2"}}, 
 {"id":"P3","film_id":"F2","computed_films_m2m_marked_as_m2o":{"id":"F4"}}, 
 {"id":"P4","film_id":"F3","computed_films_m2m_marked_as_m2o":{"id":"F1"}}, 
 {"id":"P4","film_id":"F3","computed_films_m2m_marked_as_m2o":{"id":"F2"}}, 
 {"id":"P4","film_id":"F3","computed_films_m2m_marked_as_m2o":{"id":"F4"}}]
```

laurenceisla:
Closes #570

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1174,13 +1174,13 @@ Taking advantage of overloaded functions, you can use the same function name for
     select * from directors where film_school_id = $1.id
   $$ stable language sql;
 
-.. important::
+Computed relationships have good performance as their intended design follow the `Inlining conditions for table functions <https://wiki.postgresql.org/wiki/Inlining_of_SQL_functions#Inlining_conditions_for_table_functions>`_.
 
-  - Computed relationships have good performance when they are inlined, that is, when they are created following the `inlining conditions for table functions <https://wiki.postgresql.org/wiki/Inlining_of_SQL_functions#Inlining_conditions_for_table_functions>`_.
+.. warning::
 
-  - Computed relationships can return a table without using `SETOF`, but bear in mind that they will not be inlined.
+  - Always use ``SETOF`` when creating computed relationships. Functions can return a table without using ``SETOF``, but bear in mind that they will not be inlined.
 
-  - When using the ``ROWS 1`` estimation, PostgREST will expect that a single row is returned. If that is not the case, then it will unnest the embedding and return repeated values for the top level resource.
+  - Make sure to correctly label the ``to-one`` part of the relationship. When using the ``ROWS 1`` estimation, PostgREST will expect a single row to be returned. If that is not the case, then it will unnest the embedding and return repeated values for the top level resource.
 
 .. _nested_embedding:
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1174,7 +1174,13 @@ Taking advantage of overloaded functions, you can use the same function name for
     select * from directors where film_school_id = $1.id
   $$ stable language sql;
 
-Computed relationships have good performance as they follow the `Inlining conditions for table functions <https://wiki.postgresql.org/wiki/Inlining_of_SQL_functions#Inlining_conditions_for_table_functions>`_.
+.. important::
+
+  - Computed relationships have good performance when they are inlined, that is, when they are created following the `inlining conditions for table functions <https://wiki.postgresql.org/wiki/Inlining_of_SQL_functions#Inlining_conditions_for_table_functions>`_.
+
+  - Computed relationships can return a table without using `SETOF`, but bear in mind that they will not be inlined.
+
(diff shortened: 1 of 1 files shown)


docs/api.rst
```diff
@@ -1174,7 +1174,15 @@ Taking advantage of overloaded functions, you can use the same function name for
...
   $$ stable language sql;
 
-Computed relationships have good performance as they follow the `Inlining conditions for table functions <https://wiki.postgresql.org/wiki/Inlining_of_SQL_functions#Inlining_conditions_for_table_functions>`_.
+.. important::
+
+  There are some caveats when working with computed relationships:
+
+  - Computed relationships may have good performance when they are inlined, that is, when they are created following the `inlining conditions for table functions <https://wiki.postgresql.org/wiki/Inlining_of_SQL_functions#Inlining_conditions_for_table_functions>`_.
+
+  - Currently, only the computed relationships that return a ``SETOF`` can be inlined.
```
steve-chavez, line 1183: I wonder if we should add this since it's already included in the inlining guide on the pg wiki.

laurenceisla: Hmmm, yeah you're right, the condition says:

> The function is declared RETURNS SETOF or RETURNS TABLE

And `RETURNS TABLE` does not work for computed relationships. On the other hand, `RETURNS <tablename>`  does work for computed rels, but they are not inlined.  I'll change the wording there.

docs/api.rst
```diff
@@ -1174,7 +1174,13 @@ Taking advantage of overloaded functions, you can use the same function name for
...
   $$ stable language sql;
 
-Computed relationships have good performance as they follow the `Inlining conditions for table functions <https://wiki.postgresql.org/wiki/Inlining_of_SQL_functions#Inlining_conditions_for_table_functions>`_.
+.. important::
+
+  - Computed relationships have good performance when they are inlined, that is, when they are created following the `inlining conditions for table functions <https://wiki.postgresql.org/wiki/Inlining_of_SQL_functions#Inlining_conditions_for_table_functions>`_.
+
+  - Computed relationships can return a table without using `SETOF`, but bear in mind that they will not be inlined.
+
+  - When using the ``ROWS 1`` estimation, PostgREST will expect that a single row is returned. If that is not the case, then it will unnest the embedding and return repeated values for the top level resource.
```
steve-chavez, line 1183: The planned interface for computed relationships, as it is documented now, do follow the inlining guidelines. Users will have to visit the documentation for learning about this anyway. They can go out of their way to break the recommendations, but this doesn't mean we have to undersell the feature(the way it looks to me now).

So how about this:

```suggestion
Computed relationships have good performance as their intended design follow the `Inlining conditions for table functions <https://wiki.postgresql.org/wiki/Inlining_of_SQL_functions#Inlining_conditions_for_table_functions>`_.

.. warning::

  - Make sure to always use `SETOF` when using computed relationships. Functions can return a table without using `SETOF`, but bear in mind that they will not be inlined.

  - Make sure to correctly label the `to-one` relationship end. When using the ``ROWS 1`` estimation, PostgREST will expect that a single row is returned. If that is not the case, then it will unnest the embedding and return repeated values for the top level resource.
```

PR #613, opened 2023-05-01
Authentication as a reference

steve-chavez:
* Moves Users and Groups into its own Database Authorization topic
* Moves JWT from Auth0 into Ecosystem
* Moves SQL User Management to a how-to

This is needed for documenting https://github.com/PostgREST/postgrest/pull/2742/.

## Todo

- [x] Refine wording and create links between the different pages

### DIFF
--- a/docs/auth.rst
+++ b/docs/auth.rst
@@ -1,22 +1,47 @@
-.. _roles:
+.. raw:: html
 
-Overview of Role System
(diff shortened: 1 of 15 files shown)


steve-chavez:
I think it's overall clearer. JWT is a feature and didn't have a reference, now we can document it better.

PR #639, opened 2023-06-08
Note content-type for singular objects

dbaynard:
Resolves PostgREST/postgrest#2809

I hit this issue when replacing the backend for a client that was naively testing the `Content-Type` was `application/json`. This small change would have saved me significant debugging. Hopefully it will help others, too.

### DIFF
--- a/docs/references/api/resource_representation.rst
+++ b/docs/references/api/resource_representation.rst
@@ -66,7 +66,7 @@ This returns
 
   { "id": 1 }
 
-with a :code:`Content-Type` that begins :code:`application/vnd.pgrst.object+json`.
+with a :code:`Content-Type: application/vnd.pgrst.object+json`.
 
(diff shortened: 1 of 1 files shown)


docs/references/api/resource_representation.rst
```diff
@@ -66,6 +66,8 @@ This returns
 
   { "id": 1 }
 
+with a :code:`Content-Type` that begins :code:`application/vnd.pgrst.object+json`.
```
steve-chavez, line 69: How about
```suggestion
with a :code:`Content-Type: application/vnd.pgrst.object+json`.
```

steve-chavez (approved):
Thanks for the contrib!

dbaynard:
Thank you — I don't like closing issues with no action where there's something (even something small) that can be doing to improve things for the future.

PR #649, opened 2023-06-21
Change example getallfilms to stable instead of immutable

wyozi:
Hi, while reading the docs I noticed that example function `getallfilms` was marked as immutable. However, wouldn't stable be a more correct volatility marker given the fact that it queries the database and returns value of a database select?


steve-chavez (approved):
Yes, it should be stable. Thanks for the contrib!

Issue #662, opened 2023-07-18
The table of configuration parameters is not responsive


laurenceisla:
On vertical screens, the table does not show the env/database configuration names:

![test](https://github.com/PostgREST/postgrest-docs/assets/18169764/d5e7c6a9-f94a-4300-b08f-255fc3ced6b5)

We could:
- Add a horizontal scroll to the table
- Rollback to env/database names below each configuration, e.g.: https://postgrest.org/en/v11.0/references/configuration.html#admin-server-port

steve-chavez:
@laurenceisla Good catch!

So when I put it all in the table my motivation was to DRY the configs. 

> Rollback to env/database names below each configuration, e.g.:

I believe that's the solution. So for example on https://postgrest.org/en/v11.1/references/connection_pool.html#dynamic-connection-pool, I have to inline the default value of the config. If each default is on the config section that wouldn't be needed.

So to DRY things we need to remove the table and go back to the previous layout.

PR #663, opened 2023-07-21
dry config and make it responsive

steve-chavez:
Closes https://github.com/PostgREST/postgrest-docs/issues/662


steve-chavez:
The configs now look better but I think the `List of Parameters` section needs subsections. Like `server configs`, `pool configs`, etc.

Issue #644, opened 2023-06-14
Move release notes pages to github releases


steve-chavez:
Just saw [QuestDB release page](https://github.com/questdb/questdb/releases/tag/7.2). Really cool, made me remember we have full markdown support there.

How about if we deprecate the docs release notes and move them to GH release page? https://postgrest.org/en/stable/releases/v11.1.0.html

### Advantages

- DRY. We no longer have to duplicate the CHANGELOG entries. Features can be linked to docs sections.
- Users get notifications on releases

### Notes

Seeing the QuestDB page, I realize it's better to have frequent releases. That way the release post overview can be focused.

steve-chavez:
@laurenceisla @wolfgangwalther WDYT?

laurenceisla:
I like the idea, there could be a single Release page in the docs that redirects to each of the GH releases. A thing to keep in mind is that we must add links to the docs for the relevant features, changes, etc. so we should keep the docs up to date for that reason (even if the release page can be updated later).

steve-chavez:
I believe the end result looks better: https://github.com/PostgREST/postgrest/releases/tag/v11.2.0

That being said, it was a bit of pain having to sync the docs with core. Doing https://github.com/PostgREST/postgrest/issues/2814 should make this easier.

PR #666, opened 2023-08-04
Towards v11.2.0

steve-chavez:
Closes https://github.com/PostgREST/postgrest-docs/issues/644.


laurenceisla:
Since we're reverting using computed columns for disambiguation. I made two possible alternatives:

1. Keeping the "Multiple M2O" and "Multiple O2M" examples: [25549d7](https://github.com/PostgREST/postgrest-docs/pull/666/commits/25549d7d6fbf1446f834e917829035add1cc6517)
2. Keeping the old structure using "Hint" with vews as example. But with a new title:  "Multiple FK Relationships to Many Resources": [a612db9](https://github.com/PostgREST/postgrest-docs/pull/666/commits/a612db992f883e4923dcc5a18037319cb822a2dc)

The first version is more straightforward but it's not that accurate since only the foreign key can be used for both those cases (m2o and o2m), which is simpler.

The second one is more accurate in explaining when it's needed, but not so straightforward as the previous one.

steve-chavez:
> "Multiple FK Relationships to Many Resources"

Since we're explaining FK joins as a database concern, the term "resources" doesn't seem fit. I've removed that part and restored the previous examples plus used the `hint` syntax.

PR #669, opened 2023-08-26
Add a regular account to reduce operational security risks

haowuge:
Add a regular account to reduce operational security risks, and set a failed retry 3 times.

### DIFF
--- a/docs/integrations/systemd.rst
+++ b/docs/integrations/systemd.rst
@@ -12,6 +12,12 @@ First, create postgrest configuration in ``/etc/postgrest/config``
   db-anon-role = "<your_anon_role>"
   jwt-secret = "<your_secret>"
 
+Create a dedicated  ``postgrest`` user with:
+
+.. code-block:: ini
+
+  sudo useradd -M -U -d /nonexistent -s /usr/sbin/nologin postgrest
+
 Then create the systemd service file in ``/etc/systemd/system/postgrest.service``
 
 .. code-block:: ini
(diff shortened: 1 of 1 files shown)


docs/integrations/systemd.rst
```diff
@@ -12,6 +12,12 @@ First, create postgrest configuration in ``/etc/postgrest/config``
   db-anon-role = "<your_anon_role>"
   jwt-secret = "<your_secret>"
 
+Create a dedicated  ``postgrest`` user with:
+
+.. code-block:: ini
+
+  sudo useradd -M -U -d /nonexistent -s /usr/sbin/nologin postgrest
```
steve-chavez, line 19: Cool. TIL about `/nonexistent`: https://www.debian.org/doc/debian-policy/ch-opersys.html#non-existent-home-directories

docs/integrations/systemd.rst
```diff
@@ -21,6 +27,9 @@ Then create the systemd service file in ``/etc/systemd/system/postgrest.service`
   After=postgresql.service
 
   [Service]
+  User=postgrest
+  Group=postgrest
+  Restart=on-failure:3
```
steve-chavez, line 32: @haowuge Hm, what does the 3 here mean? Could you link the docs for that?

I cannot find a reference to it on https://www.freedesktop.org/software/systemd/man/systemd.service.html#Restart=

haowuge: @steve-chavez Thank you for your guidance,
I mistakenly remembered 'Restart=on failure: 3', which is actually a parameter of the docker. 
Sorry

[Docker supports the following restart policies: ](https://docs.docker.com/engine/reference/run/)

steve-chavez (approved):
@haowuge Thanks for the contrib!

PR #680, opened 2023-09-18
explanation: data independence

steve-chavez:
WIP for the discussion we had on https://github.com/PostgREST/postgrest/issues/2310#issuecomment-1694562855.

This should explain domain representations better.

### DIFF
--- a/docs/explanations/data_independence.rst
+++ b/docs/explanations/data_independence.rst
@@ -1,0 +1,10 @@
+.. _data_independence:
+
+Data Independence
+#################
+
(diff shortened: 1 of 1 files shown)


wolfgangwalther:
For https://github.com/PostgREST/postgrest/issues/2814, I'd like to deal with all the open PRs in the docs repo.

Since this PR is not outdated much and just adds a single file, I will leave this open. This can easily be copied to the main repo later manually, when you continue to work on it. I briefly considered putting the information in here into an issue, but leaving it open is probably easier as a reference for that piece of text.

PR #686, opened 2023-10-05
Version not set when invalid URI or key/value format

laurenceisla:
Related to https://github.com/PostgREST/postgrest/pull/2985

### DIFF
--- a/docs/references/admin.rst
+++ b/docs/references/admin.rst
@@ -108,7 +108,8 @@ When debugging a problem it's important to verify the running PostgREST version.
 
 .. note::
 
-  The server sets the `fallback_application_name <https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-FALLBACK-APPLICATION-NAME>`_ for this query to work. To override the value set ``application_name`` on the connection string.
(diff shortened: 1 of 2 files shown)


docs/references/admin.rst
```diff
@@ -108,7 +108,8 @@ When debugging a problem it's important to verify the running PostgREST version.
 
 .. note::
 
-  The server sets the `fallback_application_name <https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-FALLBACK-APPLICATION-NAME>`_ for this query to work. To override the value set ``application_name`` on the connection string.
+  - The server sets the `fallback_application_name <https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-FALLBACK-APPLICATION-NAME>`_ to the connection URI for this query to work. To override the value set ``application_name`` on the connection string.
+  - The version will not be set if the connection string is in `keyword/value format <https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING-KEYWORD-VALUE>`_ or if it contains certain unescaped special characters.
```
steve-chavez, line 112: Better make this an `important` note for more visibility. Also:

```suggestion
  - The version will only be set if it's a valid URI (`RFC 3986 <https://datatracker.ietf.org/doc/html/rfc3986>`_). This means any special characters must be urlencoded.
  - The version will not be set if the connection string is in `keyword/value format <https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING-KEYWORD-VALUE>`_.
```

steve-chavez (approved):

PR #688, opened 2023-10-25
add documentation for server-cors-allowed-origins config

taimoorzaeem:
Add documentation for CORS config.

### DIFF
--- a/docs/references/api/cors.rst
+++ b/docs/references/api/cors.rst
@@ -1,7 +1,10 @@
-CORS
-====
+.. _cors:
 
-PostgREST sets highly permissive cross origin resource sharing, that is why it accepts Ajax requests from any domain.
+CORS
+####
+
+By default, PostgREST sets highly permissive cross origin resource sharing, that is why it accepts Ajax requests from any domain. This behavior can be configured by using :ref:`server_cors_allowed_origins`.
+
 
 It also handles `preflight requests <https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request>`_ done by the browser, which are cached using the returned ``Access-Control-Max-Age: 86400`` header (86400 seconds = 24 hours). This is useful to reduce the latency of the subsequent requests.
 
@@ -32,3 +35,16 @@ A ``POST`` preflight request would look like this:
   Access-Control-Allow-Methods: GET, POST, PATCH, PUT, DELETE, OPTIONS, HEAD
   Access-Control-Allow-Headers: Authorization, Content-Type, Accept, Accept-Language, Content-Language
   Access-Control-Max-Age: 86400
+
+.. _allowed_origins:
+
+Allowed Origins
+===============
+
+With the below config setting, PostgREST will accept CORS requests from domains :code:`http://example.com` and :code:`http://example2.com`.
+
+
+.. code-block::
+  
+  server-cors-allowed-origins="http://example.com, http://example2.com"
+
--- a/docs/references/api/cors.rst
+++ b/docs/references/api/cors.rst
@@ -41,7 +41,7 @@ A ``POST`` preflight request would look like this:
 Allowed Origins
 ===============
 
-With the below config setting, PostgREST will accept CORS requests from domains :code:`http://example.com` and :code:`http://example2.com`.
+With the following config setting, PostgREST will accept CORS requests from domains :code:`http://example.com` and :code:`http://example2.com`.
 
 
 .. code-block::
--- a/docs/references/configuration.rst
+++ b/docs/references/configuration.rst
@@ -774,6 +774,23 @@ raw-media-types
 
    raw-media-types="image/png, font/woff2"
 
+.. _server_cors_allowed_origins:
+
+server-cors-allowed-origins
+---------------------------
+
+  =============== ===================================
+  **Type**        String
+  **Default**     `n/a` 
+  **Reloadable**  N
+  **Environment** PGRST_SERVER_CORS_ALLOWED_ORIGINS
+  **In-Database** `pgrst.server_cors_allowed_origins`
+  =============== ===================================
+
+  Specifies allowed CORS origins in this config. See :ref:`cors`.
+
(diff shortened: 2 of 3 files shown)


docs/references/api/cors.rst
```diff
@@ -1,7 +1,18 @@
...
+PostgREST can set any cross origin resource sharing, that is why it can accept Ajax requests from any set of domains.
+
+You can specify the allowed CORS origins using the :code:`server-cors-allowed-origins` config. If the config is not set, it accepts requests from any domain. If the config is set to :code:`""`, it rejects requests from any domain. See :ref:`server_cors_allowed_origins`.
+
+With the below config setting, PostgREST will accept requests from :code:`http://example.com` and :code:`http://example2.com`.
+
+.. code-block::
+  
+  server-cors-allowed-origins="http://example.com, http://example2.com"
+
```
steve-chavez, line 15: I think we should start from the most common use case, which is accept all domains:

```suggestion
By default, PostgREST sets highly permissive cross origin resource sharing, that is why it accepts Ajax requests from any domain.This behavior can be configured by using :ref:`_server_cors_allowed_origins
```

steve-chavez: Then at the end of this page, something like:

```
Allowed Origins
===============

With the below config setting, PostgREST will accept requests from :code:`http://example.com` and :code:`http://example2.com`.


.. code-block::
  
  server-cors-allowed-origins="http://example.com, http://example2.com"

```

docs/references/api/cors.rst
```diff
@@ -1,7 +1,18 @@
+.. _cors:
+
 CORS
 ====
 
-PostgREST sets highly permissive cross origin resource sharing, that is why it accepts Ajax requests from any domain.
+PostgREST can set any cross origin resource sharing, that is why it can accept Ajax requests from any set of domains.
+
+You can specify the allowed CORS origins using the :code:`server-cors-allowed-origins` config. If the config is not set, it accepts requests from any domain. If the config is set to :code:`""`, it rejects requests from any domain. See :ref:`server_cors_allowed_origins`.
```
steve-chavez, line 8: These parts would fit better on the config page.

```
If the config is not set, it accepts requests from any domain. If the config is set to :code:`""`, it rejects requests from any domain.
```

They can be bullets like here https://postgrest--688.org.readthedocs.build/en/688/references/configuration.html#server-host

docs/references/configuration.rst
```diff
@@ -774,6 +774,24 @@ raw-media-types
...
+  **Default**     `n/a` 
+  **Reloadable**  N
+  **Environment** PGRST_SERVER_CORS_ALLOWED_ORIGINS
+  **In-Database** `pgrst.server_cors_allowed_origins`
+  =============== ===================================
+
+  Specifies allowed CORS origins in this config. See :ref:`cors`.
+
+  * When this is not set, PostgREST **accepts** CORS requests from any domain
+  * When this is set to :code:`""`, PostgREST **rejects** CORS requests from any domain
```
steve-chavez, line 793: @taimoorzaeem I wonder if this is useful behavior. Wouldn't be better to treat `""` the same as not set - accept all reqs from any domain?

laurenceisla: Ah, that was my suggestion in https://github.com/PostgREST/postgrest/pull/2986#pullrequestreview-1694030855. It's not so much that rejects any request, i.e. the ones with empty "Origin: " will work (kinda impossible case). In other words, when this is set then the values will be used as-is. The note could be: "To allow all origins do not set this config, otherwise, all the values inside will be used as-is".

On the other hand, I get that it may not be too intuitive, maybe we could add this special case for "" or even "*" (which will be invalid if the config has more values for any reason "*,http://google.com,...").

steve-chavez: @laurenceisla Got it. As you mentioned is like an empty search path. But it does look "sloppy" and perhaps some user could interpolate an empty value `..="$EMPTY"` in the confg and be left surprised as to why all requests are failing?

I believe we should default `""` to accept all. WDYT?

laurenceisla: A quick look on the `db-extra-search-path` returned `[""]` when empty. But it does not give any issue if that's the case.

> I believe we should default "" to accept all. WDYT?

Yes, agree. Treat only an empty string as if it wasn't set at all. Any other value, would be considered as-is.

laurenceisla: > IIRC we did the same for many other configs "" translates to Nothing?

Funny that I tried setting `db-schemas = ""` to test this functionally,  and it worked OK when requesting tables, because it defaulted to the `public` schema (as if it wasn't set at all). Then, while working on another issue with RPCs, it gave me this error:

`message: "Could not find the function .index without parameters in the schema cache"`

Had me scratching my head for a while, then remembered the empty setting. So it looks like it's not defaulting to `public` in the RPC search... odd.

laurenceisla: I implemented this in #3028. Now the last line can be removed, and the info on the empty strings should be added to the first one.

docs/references/api/cors.rst
```diff
@@ -32,3 +35,16 @@ A ``POST`` preflight request would look like this:
   Access-Control-Allow-Methods: GET, POST, PATCH, PUT, DELETE, OPTIONS, HEAD
   Access-Control-Allow-Headers: Authorization, Content-Type, Accept, Accept-Language, Content-Language
   Access-Control-Max-Age: 86400
+
+.. _allowed_origins:
+
+Allowed Origins
+===============
+
+With the below config setting, PostgREST will accept CORS requests from domains :code:`http://example.com` and :code:`http://example2.com`.
```
laurenceisla, line 44:
```suggestion
With the following config setting, PostgREST will accept CORS requests from domains :code:`http://example.com` and :code:`http://example2.com`.
```

TIL that "bellow" [is not commonly used as an adjective](https://english.stackexchange.com/a/611).

laurenceisla (approved):

steve-chavez (approved):

PR #691, opened 2023-11-03
Update providing-html-content-using-htmx.rst

joaofelipecb:
Included information about configuration of raw-media-types. Without this configuration the example doesn't work correctly.

### DIFF
--- a/docs/how-tos/providing-html-content-using-htmx.rst
+++ b/docs/how-tos/providing-html-content-using-htmx.rst
@@ -30,6 +30,13 @@ and return a raw HTML document file.
 
   create domain "text/html" as text;
 
+It will be needed to configure the property raw-media-types (https://postgrest.org/en/latest/references/configuration.html#raw-media-types) as well.
+
+.. code-block:: postgres
+
+  raw-media-types="text/html"
+
+
 Creating an HTML Response
 -------------------------
 


steve-chavez:
Thanks @joaofelipecb.

We're actually trying to improve the DX of that tutorial, so we've removed `raw-media-types` on the master branch of postgrest (https://github.com/PostgREST/postgrest/pull/2825) in preference of the DOMAIN media types. Note that this are the `latest` docs.

How about if instead we put a warning mentioning that the tutorial only works on the master branch and perhaps link a nightly release there?

joaofelipecb:
I understand the worry with DX and in the future the nightly version will become stable so the tutorial will work in a stable version.

Unfortunately, I spend 2h to figure out why this was not working, and despite the high quality of the project, this harm my own developer experience, since I almost quit the tutorial.

Since I want more developers to try PostgRest, I think the better solution is to provide a troubleshooting section and insert the error message of prior versions and the solution.

steve-chavez:
> Unfortunately, I spend 2h to figure out why this was not working, and despite the high quality of the project, this harmed my own developer experience, as I almost quit the tutorial.

Sorry about that @joaofelipecb. We'll definitely be more careful about publishing how-tos that rely on master from now on.

> Since I want more developers to try PostgRest, I think the better solution is to provide a troubleshooting section and insert the error message of prior versions and the solution. What do you think?

How about doing having tabs on the code snippets? Like we do for HTTP/Curl examples on https://postgrest.org/en/latest/references/api/tables_views.html).

For this we could have a tab for `v11 and previous` and `latest` (which later will become v12). What do you think?

steve-chavez:
Now that v12 is released the tutorial is stable: https://postgrest.org/en/stable/how-tos/providing-html-content-using-htmx.html

PR #697, opened 2023-11-14
make the jwt generation easier :)

anvarknian:

### DIFF
--- a/docs/tutorials/tut1.rst
+++ b/docs/tutorials/tut1.rst
@@ -49,7 +49,6 @@ check the :code:`tutorial.conf` (created in the previous tutorial) the line with
 .. code-block:: ini
 
   # PASSWORD MUST BE AT LEAST 32 CHARS LONG
-  # add this line to tutorial.conf:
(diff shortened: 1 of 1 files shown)


docs/tutorials/tut1.rst
```diff
@@ -41,16 +41,15 @@ Let's create a password and provide it to PostgREST. Think of a nice long one, o
     export LC_CTYPE=C
 
     # read random bytes and keep only alphanumerics
-    < /dev/urandom tr -dc A-Za-z0-9 | head -c32
+    echo "jwt-secret = \"$(< /dev/urandom tr -dc A-Za-z0-9 | head -c32)\"" > tutorial.conf
```
steve-chavez, line 44: `>` is the replace command. Changing it to `>>` (append) instead. Otherwise it'll override the tutorial.conf from tutorial 0.

```suggestion
    echo "jwt-secret = \"$(< /dev/urandom tr -dc A-Za-z0-9 | head -c32)\"" >> tutorial.conf
```

PR #727, opened 2023-12-19
Add warning and fix to htmx how-to

laurenceisla:
Adds experimental warning and fix html sanitization.

### DIFF
--- a/docs/how-tos/providing-html-content-using-htmx.rst
+++ b/docs/how-tos/providing-html-content-using-htmx.rst
@@ -11,6 +11,11 @@ Htmx expects an HTML response and uses it to replace an element inside the DOM (
 
 .. image:: ../_static/how-tos/htmx-demo.gif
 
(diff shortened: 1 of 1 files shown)


docs/how-tos/providing-html-content-using-htmx.rst
```diff
@@ -11,6 +11,11 @@ Htmx expects an HTML response and uses it to replace an element inside the DOM (
 
 .. image:: ../_static/how-tos/htmx-demo.gif
 
+.. warning::
+
+  This is a proof of concept showing what can be achieved using both technologies.
+  Some security concerns may be overlooked. Use at your own discretion.
```
steve-chavez, line 17:
```suggestion
  This is a proof of concept showing what can be achieved using both technologies.
  We are working on `plmustache <https://github.com/PostgREST/plmustache>`_ which will further improve the...
```

PR #731, opened 2023-12-31
Fix secret command to avoid non-ascii alphanumerics

mrarticuno:
Example of result with the older command:
`��l6N�z���ݹ�raDp��yQMd1���Zah�`


steve-chavez (approved):
Thanks!

PR #739, opened 2024-01-15
Fix Sphinx version error in RTD

laurenceisla:
Fixing the "Sphinx version error" in RTD:

```log
Sphinx version error:
The sphinxcontrib.applehelp extension used by this project needs at least Sphinx v5.0; it therefore cannot be built with this version.
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 docutils==0.17.1
-sphinx==5.0.2
+sphinx>=5.0.2
 sphinx-copybutton
 sphinx-rtd-theme>=0.5.1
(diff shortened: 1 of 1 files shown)


laurenceisla:
In summary I bumped docutils, sphinx and sphinx-tabs. Docutils gave formating errors in #455, IIRC it broke lists which are looking OK now, so it looks that it was fixed. Unless it was another formatting problem, then this should be ready to merge.

steve-chavez (approved):

.test
wolfgangwalther, line 1: @laurenceisla just stumbled over this empty file that you committed here. I can safely remove that, right?

PR #740, opened 2024-01-15
Spelling correction

Parashoe:
Spell correct "callounter" to "callcounter"


steve-chavez (approved):
Thanks for the contrib!

